PR #39, opened 2014-07-01
Trap Async DNS Resolution errors and throw an exception

ck99:
Added some handlers to throw exceptions during Async DNS Resolution.


kelunik:
Closing due to inactivity and merge conflicts.

Issue #65, opened 2014-11-04
Feature Request - FormBody() methods addFields() and addFiles() for arrays


FabienO:
Assuming this is the correct place for this.

Can we please have methods to support adding fields to FormBody() as an array please.

Example:
```
FormBody::addFields(array $associativeFieldMap); 
FormBody::addFiles(array $arr);
```

PR #72, opened 2015-01-20
Add FormBody::addFields and FormBody::addFiles

kelunik:
This is an implementation for issue #65. It adds the ability to set multiple fields / files at once by using an array.


rdlowrey:
:+1:

PR #76, opened 2015-03-17
Added type checks for request setters

kelunik:


staabm:
Shouldnt it be IllegalArgumentExceptioninstead?

kelunik:
It's `InvalidArgumentException` in PHP, but yes, probably. But we're using `DomainException` everywhere else, so I also used it here, see https://github.com/amphp/artax/search?utf8=%E2%9C%93&q=domainexception

kelunik:
Closing, as this would break passing e.g. `Uri` objects.

PR #78, opened 2015-04-12
Allow setting default user agent option

cspray:
Adds support for setting a default user agent to override the Artax UA.

The reason for this is so that API clients can specify the library sending requests instead of the underlying HTTP client doing the "leg work". This would provide an easy way for these clients to set the UA on every request it sends.

I didn't provide a test because there were slim pickins' in the existing test. I wouldn't mind writing a test for this but would probably need some guidance on the best way to approach that.


kelunik:
Thanks!

PR #84, opened 2015-07-28
no Amp\wait or run methods. Fix the documentation

harikt:
Hi, 

It has came to my notice from (https://github.com/amphp/artax/issues/83#issuecomment-125572351) that the documentation is wrong. It don't have `wait` or `run` methods for `Amp` and the namespace is also not there for the `Client` . My assumption is @staabm would have shown synchronous one itself for the documentation and we may need to use AsyncClient for non blocking one or I am still not sure of the same.

Thank you.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,9 +51,8 @@ as the request parameter:
 <?php
 
 try {
-    $client = new Amp\Artax\Client;
-    $promise = $client->request('http://www.google.com');
-    $response = Amp\wait($promise);
+    $client = new Artax\Client;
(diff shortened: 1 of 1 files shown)


rdlowrey:
I'm sorry for the service interruption here -- we've been working hard to finish up and release amp 1.0.0. Now that that is finished I should be able to fix the breakage here ASAP.

kelunik:
Closing this, as `Amp\wait` is present in Amp v1.0.0, please use the `dev-v2` for now.

PR #86, opened 2015-09-09
Fix improper type parser

cspray:
Horrible branch name not withstanding...

- Removes erroneous whitespace-only lines in Uri
- It is possible in PHP7 with strict types declared to pass null to `http_build_query` resulting in an error. Demonstrated here: https://3v4l.org/M0pJO  This change ensures that a string is always passed


cspray:
Sorry, I submitted these with the incorrect email configured. I'll be closing this PR and reopening with correct git configuration.

Issue #73, opened 2015-02-04
Support proxy authentication


rdlowrey:
Currently there is no means to support `Proxy-Authorization:` headers ... needs to be fixed :)

Bilge:
Confirmed, also needs supporting in `HTTP_PROXY`/`HTTPS_PROXY` environment fields.

kelunik:
Those are already supported, but only in CLI mode, because Httpoxy.

Bilge:
They do not support authentication.

kelunik:
Built-in proxy support will probably not be part of 4.0 and can be implemented independently via the `Connector` interface.

Bilge:
Do you mean proxy *authentication* will not be part of 4.0, or are you completely stripping out all proxy support in 4.0?

kelunik:
We completely removed it for now, can be reimplemented using a `Connector`.

Bilge:
You could completely remove the library and it be reimplemented by coding everything from scratch, but that assurance doesn't grant much comfort. Though I don't know your reasons for removing proxy support, it seems an incorrect decision.

kelunik:
I don't care about proxy support and the previous implementation had a very sub-optimal API compared to the new socket Connector interface and didn't have any tests. Things that are more important to me are e.g. HTTP/2 and caching.

If you care for HTTP connect proxy support, please open a PR to the socket library turning the previous HttpTunnler into a Connector and include tests for it.

PR #89, opened 2015-12-22
Basic proxy auth

lt:
By basic, I mean minimum viable for basic digest auth.

Does not handle 407 headers, and does not upgrade the underlying proxy framework to be able to use SSL, so passwords are sent in the clear.

Let's call it progress towards #73, but wouldn't call this finished product quality by a long shot.

### DIFF
--- a/lib/HttpSocketPool.php
+++ b/lib/HttpSocketPool.php
@@ -8,7 +8,7 @@ use Amp\Deferred;
 class HttpSocketPool {
     const OP_PROXY_HTTP = 'op.proxy-http';
     const OP_PROXY_HTTPS = 'op.proxy-https';
-    const OP_PROXY_HTTP_AUTH = 'op.proxy-https-auth';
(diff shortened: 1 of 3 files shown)


lib/HttpSocketPool.php
```diff
@@ -8,12 +8,16 @@
 class HttpSocketPool {
     const OP_PROXY_HTTP = 'op.proxy-http';
     const OP_PROXY_HTTPS = 'op.proxy-https';
+    const OP_PROXY_HTTP_AUTH = 'op.proxy-https-auth';
```
kelunik: `s/https/http`.

lib/HttpTunneler.php
```diff
@@ -14,11 +14,12 @@ class HttpTunneler {
      * @param string $authority
      * @return \Amp\Promise
      */
-    public function tunnel($socket, $authority) {
+    public function tunnel($socket, $authority, $proxyAuth) {
```
kelunik: `$proxyAuth` should probably have a default value for BC reasons.

PR #91, opened 2016-02-07
Exposes relevant parser options to client

cspray:
The primary use case for this PR is being able to set the `Parser::OP_MAX_BODY_BYTES` option for consuming streaming connections (e.g. the Twitter streaming API). This patch ensures that when handling these types of connections we can configure the Parser appropriately to not check body size. Otherwise after the data limit is hit an error is thrown and the connection is lost.

The rest of the options were added as they seemed relevant (there was a TODO for the body swap size) and it made sense to hit them all at once.


kelunik:
LGTM, @bwoebi?

bwoebi:
@kelunik yep, feel free to merge :-)

kelunik:
Thanks, @cspray!

PR #107, opened 2017-04-06
Fixed Cleint::enableCrypto method

Fedott:


bwoebi:
Seems we forgot to update artax after the rename; thanks for the PR :-)

PR #108, opened 2017-04-07
Check if idleWatcher exists before disabling loop

ck99:
Was getting: 

`TypeError: Argument 1 passed to Amp\Loop::disable() must be of the type string, null given`

not sure why idleWatcher was null, so perhaps there is some other issue that needs fixing, but I did notice that other parts of SocketPool do check the status of idleWatcher before passing it to Loop operations.


bwoebi:
This looks fine to me - it's just that a null watcher was fine to be disabled in amp v1. Merging.

PR #110, opened 2017-04-12
Check for NULLs before Loop::cancel

ck99:


bwoebi:
Yep, here it looks fine; merging :-)

PR #126, opened 2017-07-10
Enforce size limit also on gzipped streams

kelunik:
Might not be the prettiest implementation, but it works.


bwoebi:
This is though only meaningful if the incoming data isn't specifically crafted to produce enormous amounts of data, right?

kelunik:
Aborts just fine with https://blog.haschek.at/tools/bomb.php

bwoebi:
Ah, yeah. Inflate only may decompress to a size maximally 1032 times larger than the input.
Given that we have a small chunk size of just 16 KB, we won't get more than 16 MB of inflated input.

So, yes, it should probably be enough.

kelunik:
Merging and releasing a bugfix version then?

PR #134, opened 2017-10-02
Allow null content type for multipart form field

brstgt:
Other clients like curl + guzzle don't send a content type for simple multipart form fields. I don't see a reason why a content type is enforced.

### DIFF
--- a/lib/FormBody.php
+++ b/lib/FormBody.php
@@ -33,7 +33,7 @@ final class FormBody implements RequestBody {
      * @param string $value
      * @param string $contentType
      */
-    public function addField(string $name, string $value, string $contentType = '') {
+    public function addField(string $name, string $value, string $contentType = 'text/plain') {
         $this->fields[] = [$name, $value, $contentType, null];
         $this->resetCache();
     }
@@ -44,7 +44,7 @@ final class FormBody implements RequestBody {
      * @param array  $data
      * @param string $contentType
      */
-    public function addFields(array $data, string $contentType = '') {
+    public function addFields(array $data, string $contentType = 'text/plain') {
         foreach ($data as $key => $value) {
             $this->addField($key, $value, $contentType);
         }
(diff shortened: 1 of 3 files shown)


kelunik:
If there's no content type, `text/plain` is assumed by the other party anyway.

brstgt:
Sure, but why transfer if it is default? It's nits, I just stumbled across it during tests + comparison with other popular clients.

bwoebi:
Is there any disadvantage in being explicit? _If_ you already use a bloated multipart form field instead of urlencoded data, it doesn't really harm to just send the default content-type.

brstgt:
As already mentioned, these are nits. You can close the PR if you don't like it.

morrisonlevi:
> Any HTTP/1.1 message containing an entity-body SHOULD include a
Content-Type header field defining the media type of that body. If
and only if the media type is not given by a Content-Type field, the
recipient MAY attempt to guess the media type via inspection of its
content and/or the name extension(s) of the URI used to identify the
resource. If the media type remains unknown, the recipient SHOULD
treat it as type "application/octet-stream".

This means:

  1. A `Content-Type` header is not required.
  2. When missing it does not imply `text/plain`. Our default to `text/plain` is therefore an error. When missing the content type should be omitted.

lib/FormBody.php
```diff
@@ -122,9 +122,14 @@ private function generateMultipartFileHeader(string $name, string $fileName, str
         return $header;
     }
 
-    private function generateMultipartFieldHeader(string $name, string $contentType): string {
+    private function generateMultipartFieldHeader(string $name, ?string $contentType): string {
         $header = "Content-Disposition: form-data; name=\"{$name}\"\r\n";
-        $header .= "Content-Type: {$contentType}\r\n\r\n";
+        if ($contentType !== null) {
```
kelunik: I think we can use `""` (empty string) instead of nullability here.

lib/FormBody.php
```diff
@@ -122,9 +122,14 @@ private function generateMultipartFileHeader(string $name, string $fileName, str
...
+	        $header .= "Content-Type: {$contentType}\r\n\r\n";
+        }
+        else {
```
kelunik: This should be on the same line as `}`.

PR #135, opened 2017-10-02
Multipart body must end with \r\n

brstgt:
This is RFC + otherwise aerys cuts of the last 2 bytes of the last multipart field


kelunik:
Thanks!

PR #136, opened 2017-10-02
Avoid single-char vars due to a bug in PHP 7.1.9

brstgt:
Yes, I am serious. Really. Despite that, avoiding one-char-vars increases code readability

[E_NOTICE] Undefined variable: m in /www/jaumo/vendor/amphp/artax/lib/Internal/Parser.php (Line 184)

Renaming the vars fixes the issue. Don't ask me why but we REALLY have that problem in production!

### DIFF
--- a/lib/Internal/Parser.php
+++ b/lib/Internal/Parser.php
@@ -179,10 +179,10 @@ final class Parser {
         }
 
         status_line_and_headers: {
-            if (preg_match(self::STATUS_LINE_PATTERN, $startLine, $m)) {
-                $this->protocol = $m['protocol'];
-                $this->responseCode = (int) $m['status'];
-                $this->responseReason = trim($m['reason']);
+            if (preg_match(self::STATUS_LINE_PATTERN, $startLine, $matches)) {
+                $this->protocol = $matches['protocol'];
+                $this->responseCode = (int) $matches['status'];
+                $this->responseReason = trim($matches['reason']);
             } else {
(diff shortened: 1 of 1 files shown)


kelunik:
Could you rename it to `$match`?

brstgt:
I could, but even the function signature calls it $matches ...

```
function preg_match ($pattern, $subject, array &$matches = null, $flags = 0, $offset = 0) {}
```

kelunik:
But it's always only a single match with multiple groups, no? `preg_match_all` has multiple matches. I don't care too much. Did you find a corresponding bug report on bugs.php.net?

brstgt:
Probably you could argue if this is a single match or multiple matches. I
simply oriented myself at the signature of the called method.

Bug: No I didn't find one nor filed one. But I observed it in many
different places since running Aerys. It only happens after Aerys runs for
a while, looks like a bug somewhere in the PHP runtime. Can't tell
where/when this exactly happens. It probably will never pop up in FPM.

2017-10-02 21:45 GMT+02:00 Niklas Keller <notifications@github.com>:

kelunik:
@bwoebi Could you have a look at the underlying bug or know about it?

nikic:
Do you have opcache enabled?

brstgt:
Nope, OpCache is disabled here:

Zend OPcache
--
  |  
  | Opcode Caching => Disabled
  | Optimization => Disabled
  | SHM Cache => Enabled
  | File Cache => Disabled
  | Startup Failed => Opcode Caching is disabled for CLI

PR #137, opened 2017-10-05
Allow empty content type for multipart form field

brstgt:
Sorry for the new PR. I squashed the changes + created a new branch

### DIFF
--- a/lib/FormBody.php
+++ b/lib/FormBody.php
@@ -124,7 +124,11 @@ final class FormBody implements RequestBody {
 
     private function generateMultipartFieldHeader(string $name, string $contentType): string {
         $header = "Content-Disposition: form-data; name=\"{$name}\"\r\n";
-        $header .= "Content-Type: {$contentType}\r\n\r\n";
+        if (!empty($contentType)) {
+            $header .= "Content-Type: {$contentType}\r\n\r\n";
+        } else {
+            $header .= "\r\n";
+        }
 
         return $header;
     }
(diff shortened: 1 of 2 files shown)


lib/FormBody.php
```diff
@@ -124,7 +124,11 @@ private function generateMultipartFileHeader(string $name, string $fileName, str
 
     private function generateMultipartFieldHeader(string $name, string $contentType): string {
         $header = "Content-Disposition: form-data; name=\"{$name}\"\r\n";
-        $header .= "Content-Type: {$contentType}\r\n\r\n";
+        if (!empty($contentType)) {
```
kelunik: Doesn't really matter here, because `0` isn't a valid content type, but please change it to `$contentType !== ""`, see https://3v4l.org/pF4sm.

test/FormBodyTest.php
```diff
@@ -0,0 +1,82 @@
...
+        $body->addField('b', 'b', 'application/json');
+        $body->addField('c', 'c', '');
+        $body->addField('d', 'd');
+        $body->addFields([
+            'e' => 'e',
+        ], '');
+        $body->addFields([
+            'f' => 'f'
+        ]);
+        $content = wait($body->createBodyStream()->read());
```
kelunik: I think this should be `$content = wait(new Message($body->createBodyStream()));`

brstgt:
Fixed

kelunik:
Thanks!

PR #144, opened 2017-10-16
Check against correct state for parser

joelwurtz:
All tests on `php-http/artax-adapter` fail since this check was added, and IMO it was the wrong constant used when reading the explanation and the code before.

However i may be wrong but this resolve tests failing on `php-http/artax-adapter` library. 

See : https://travis-ci.org/php-http/artax-adapter/builds/288221161

### DIFF
--- a/lib/DefaultClient.php
+++ b/lib/DefaultClient.php
@@ -289,7 +289,7 @@ final class DefaultClient implements Client {
                         }
                     } while (null !== $chunk = yield $socket->read());
 
-                    if ($parser->getState() !== Parser::AWAITING_HEADERS) {
+                    if ($parser->getState() !== Parser::BODY_IDENTITY_EOF) {
                         throw new HttpException("Incomplete body received.");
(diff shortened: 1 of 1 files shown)


kelunik:
If the response is complete, it should always be in `AWAITING_HEADERS` according to that:

https://github.com/amphp/artax/blob/b06769080a6a3695cf13865225da60a7a2f75b8c/lib/Internal/Parser.php#L308-L329

kelunik:
The issue is probably HTTP/1.0 without a `content-length` and without `chunked` encoding. If the response is an HTTP/1.0 response and doesn't carry any of these, we should just ignore the check / allow also `BODY_IDENTIFY_EOF`.

kelunik:
Should be fixed by 8a1cabc4a324d7039b721a9f3bda4419c483a4ce, please test.

kelunik:
Already properly solved in `master`.

PR #152, opened 2017-12-10
Allow unix sockets to be used, fixes #151

### LINKED ISSUE #151: Support unix socket
It works fine when forcing the url into the code. However there is no API to arrive there, since all the parsing stuff will throw an error when using something like `unix:///var/run/program.sock`.

Will gladly do a PR but i don't know which approach should be better :

  * Allow parsing stuff to handle unix scheme ? (will be hard IMO and add a lot of code to reproduce behavior of some PHP functions)
  * Add an extra option, that allows to override the uri of a request (like a remote_socket_endpoint option), which is not parsed (find this better, but may not be the philosophy of this package)

kelunik:
We can probably add a `FixedSocketPool` or something like that to `amphp/socket`.

### DIFF
--- a/examples/7-unix-sockets.php
+++ b/examples/7-unix-sockets.php
@@ -1,0 +1,58 @@
+<?php
+
+use Amp\Artax\HttpSocketPool;
+use Amp\Artax\Request;
+use Amp\Artax\Response;
+use Amp\CancellationToken;
+use Amp\Loop;
+use Amp\Promise;
+use Amp\Socket\BasicSocketPool;
+use Amp\Socket\ClientSocket;
+use Amp\Socket\SocketPool;
+
+require __DIR__ . '/../vendor/autoload.php';
+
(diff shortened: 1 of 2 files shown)


joelwurtz:
👍 for the fixed socket pool but can be done later, patch to the local address is on the contrary mandatory for this.

Also forget you already told me about this, my bad :/

kelunik:
@rdlowrey @bwoebi @trowski Opinions for the value to use here? I think `""` is reasonable.

trowski:
Should this be fixed here or perhaps return "unix-socket" in amphp/socket?

joelwurtz:
In this case the local adress is the same as the remote adress (both are the socket file) ?

kelunik:
That sounds reasonable as well!

kelunik:
Closing, as this will be fixed in `amphp/socket` instead.

PR #166, opened 2018-04-16
Changed DefaultClient to not delete any existing Cookie header when cookie jar doesn't overwrite it

Bilge:
There doesn't seem to be any compelling reason why, if a user has explicitly set a `Cookie` header they want to use, for the client to then delete it just because it's not also present in the "cookie jar". The cookie jar will still overwrite any existing header if it has a matching definition; this just avoids deleting the header when the cookie jar does not have any matching definition.

### DIFF
--- a/lib/DefaultClient.php
+++ b/lib/DefaultClient.php
@@ -665,7 +665,7 @@ final class DefaultClient implements Client {
 
         if (!$applicableCookies = $this->cookieJar->get($domain, $path)) {
             // No cookies matched our request; we're finished.
-            return $request->withoutHeader("Cookie");
+            return $request;
         }
 
         $isRequestSecure = strcasecmp($uri->getScheme(), "https") === 0;
@@ -682,7 +682,7 @@ final class DefaultClient implements Client {
             return $request->withHeader("Cookie", \implode("; ", $cookiePairs));
         }
 
-        return $request->withoutHeader("Cookie");
+        return $request;
     }
 
     private function generateAuthorityFromUri(Uri $uri): string {


Bilge:
As far as I know, the CI failure has nothing to do with this PR.

kelunik:
I guess the cookies should just be merged with the header values?

Bilge:
I suppose that would be ideal :+1:

kelunik:
Yes, we already have a parser in https://github.com/amphp/http.

irazasyed:
Please merge this PR if possible. I was trying to manually pass the cookies and every time I did, it never worked and then realized it was discarding them which doesn't make sense TBH.

So it would be great if the default behavior supports both ways and not just assume it.

@kelunik Please consider!

kelunik:
I will happily merge a PR that implements cookie merging, the current PR is half baked, because it has the same problem as the problem described in this PR once there is at least one cookie in the jar.

The previous header doesn't have to be parsed, the new values just need to be appended, I think.

kelunik:
@irazasyed @Bilge Please re-open against `amphp/artax` with cookie merging if you're still interested in this feature.

We've renamed `amphp/artax` to `amphp/http-client` and will continue development under the new name starting with 4.x. All changes to 3.x will happen in the old repository.

Bilge:
I still don't understand why you closed this. If this repository has been renamed, and so is still the current version of the project, why not keep the PR open? Why do you want me to open against the old fork?

kelunik:
@Bilge Because I will rework cookie handling in 4.x anyway, so the change only needs to happen for 3.x, and that needs to happen in the old repo.

PR #169, opened 2018-04-23
Fix unit-test failure in client integration test

bzikarsky:
Currently the test-suite fails since the "server closes socket after accept"-test assumes that results in an error when the response is read. In reality byte-stream already fails on write.

This adds a `catch`-clause for the byte-stream exception, converts it to an artax SocketException and changes the unit-test to expect the correct exception message.

### DIFF
--- a/lib/DefaultClient.php
+++ b/lib/DefaultClient.php
@@ -13,6 +13,7 @@ use Amp\Artax\Internal\RequestCycle;
 use Amp\ByteStream\InputStream;
 use Amp\ByteStream\IteratorStream;
 use Amp\ByteStream\Message;
+use Amp\ByteStream\StreamException;
 use Amp\ByteStream\ZlibInputStream;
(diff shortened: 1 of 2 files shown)


lib/DefaultClient.php
```diff
@@ -491,6 +492,8 @@ private function doWrite(RequestCycle $requestCycle) {
             }
 
             yield from $this->doRead($requestCycle, $socket, $connectionInfo);
+        } catch (StreamException $e) {
+            throw new SocketException("Failed to write to byte-stream", 0, $e);
```
kelunik: The catch is around the `write` and `doRead`, should the message really say `write`?

bzikarsky: Duh, probably not. I'll fix this later, I'm currently on mobile only.

kelunik:
Thanks!

PR #176, opened 2018-11-24
allow objects with __toString method as withHeaders array value

p7g:
Not sure how common of a use case this is, but it is an issue I ran into.

All tests pass, except for the one mentioned in #169.


kelunik:
LGTM.

kelunik:
Thanks!

PR #178, opened 2018-12-20
Add FormBody::getFields()

trowski:
Needed for generating OAuth 1.0 signatures.

### DIFF
--- a/lib/FormBody.php
+++ b/lib/FormBody.php
@@ -76,6 +76,17 @@ final class FormBody implements RequestBody {
         }
     }
 
+    /**
+     * Returns an array of fields, each being an array of [name, value, content-type, file-name|null].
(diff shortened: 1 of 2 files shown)


lib/FormBody.php
```diff
@@ -175,16 +190,20 @@ private function determineContentType() {
     }
 
     public function getBodyLength(): Promise {
+        if ($this->cachedLength) {
+            return $this->cachedLength;
+        }
```
kelunik: This needs to be reset somewhere.

trowski: These were reset in `resetCache()`, but never used anywhere. Leftover from refactoring?

kelunik: Maybe.

trowski:
Good to merge then?

PR #182, opened 2019-06-17
[WIP] Connection Pool Refactor

trowski:
Currently a WIP. I still need to refine the structure a bit more and add some needed functionality to how the pool handles returning connections (e.g.: keep-alive vs. close connections).


src/ClientBuilder.php
```diff
@@ -3,18 +3,18 @@
...
-    private $socketPool;
+    private $connector;
     private $networkInterceptors = [];
     private $applicationInterceptors = [];
 
-    public function __construct(?SocketPool $socketPool = null)
+    public function __construct(?Connector $connector = null)
     {
-        $this->socketPool = $socketPool ?? new UnlimitedSocketPool;
+        $this->connector = $connector ?? new DnsConnector;
```
kelunik: This should fallback to `connector()` instead.

src/Connection/DefaultConnectionPool.php
```diff
@@ -0,0 +1,41 @@
...
+use Amp\Socket\Socket;
+
+final class DefaultConnectionPool implements ConnectionPool
+{
+    private $connections = [];
+
+    public function createConnection(Socket $socket): Connection
+    {
+        $connection = new Http1Connection($socket);
+        $this->connections[$socket->getRemoteAddress()->toString()] = $connection;
```
kelunik: I wanted to say this only allows one concurrent connection per host, but actually this works, because it has the port in it.

trowski: So it allows one concurrent connection per host per port, which isn't really enough for HTTP/1.x, but works fine for HTTP/2. This is one of the areas where I'm not exactly sure how I want to approach the design.

kelunik: Ugh, well, right. I think I'd add `Connection::isBusy` and `ConnectionPool` chooses the next connection then or creates a new one if none is available. This also allows limiting the number of requests per HTTP/2 connection then, so the connection manager creates a separate HTTP/2 connection at some point.

The major issue here is that we can't use the remote address as key, because different sites might be hosted on the same IP, but due to SNI there might be different TLS certificates served.

kelunik: Ugh, well, right. I think I'd add `Connection::isBusy` and `ConnectionPool` chooses the next connection then or creates a new one if none is available. This also allows limiting the number of requests per HTTP/2 connection then, so the connection manager creates a separate HTTP/2 connection at some point.

The major issue here is that we can't use the remote address as key, because different sites might be hosted on the same IP, but due to SNI there might be different TLS certificates served.

src/Connection/DefaultConnectionPool.php
```diff
@@ -2,40 +2,107 @@
...
+    public function getConnection(Request $request, CancellationToken $cancellation): Promise
     {
-        $uri = $request->getUri();
+        return call(function () use ($request, $cancellation) {
+            $isHttps = $request->getUri()->getScheme() === 'https';
+            $defaultPort = $isHttps ? 443 : 80;
 
-        $address = $uri->getHost();
-        $port = $uri->getPort();
+            $authority = $request->getUri()->getHost() . ':' . ($request->getUri()->getPort() ?: $defaultPort);
```
kelunik: We should probably error here if we didn't get an absolute URI.

src/Connection/DefaultConnectionPool.php
```diff
@@ -2,40 +2,107 @@
...
 
-        if ($port !== null) {
-            $address .= ':' . $port;
-        }
+            if (isset($this->connections[$authority])) {
```
kelunik: We should also include the scheme in the key here, there are very rare cases where requests with `http` and `https` are supported on the same port.

src/Connection/Http1Connection.php
```diff
@@ -87,7 +95,11 @@ public function request(Request $request, ?CancellationToken $cancellation = nul
             try {
                 yield RequestWriter::writeRequest($this->socket, $request, $protocolVersion);
 
-                return yield from $this->doRead($request, $cancellation, $readingCancellation, $completionDeferred);
+                $response = yield from $this->doRead($request, $cancellation, $readingCancellation, $completionDeferred);
+
+                $this->busy = false;
```
kelunik: This is too early. You need to subscribe to the completion promise in `Response` instead, because at this point we have only read the header, not the body.

trowski: This was intentional, since another request can be made while a prior response body is being streamed… but that could cause another request to be severely delayed, so I'll use the completion promise instead.

kelunik: As far as I know no browser supports pipelining, so we probably shouldn't either.

trowski: We do in `http-server` (because we should to comply with the RFC I think…), but in the client, I agree.

src/Response.php
```diff
@@ -33,7 +36,7 @@ public function __construct(
         array $headers,
         InputStream $body,
         Request $request,
-        ConnectionInfo $connectionInfo,
+        Connection $connection,
```
kelunik: Response shouldn't have a reference to the connection itself.

trowski: I suppose not… I was hoping to just drop `ConnectionInfo`, but maybe I'll have to keep it after all. Does it make sense to pass a `Connection` to `NetworkInterceptor` too?

src/Connection/Http1Connection.php
kelunik: Yes, for `http-server` we need to support it.

src/Response.php
kelunik: `NetworkInterceptor` can receive `Connection`, that's fine. We'll have to keep `ConnectionInfo` for `Response` or remove support for it in `Response` and if you want to access it, you have to implement a `NetworkInterceptor`.

trowski: I think I like the second option, since 99.9% of the time you don't care about that information.

src/Connection/DefaultConnectionPool.php
```diff
@@ -36,73 +36,106 @@ public function __construct(?Connector $connector = null)
...
+
+            $promise->onResolve(function (?\Throwable $exception, ?Connection $connection) use ($key, $index): void {
+                if ($exception) {
+                    unset($this->connections[$key][$index]);
+
+                    if (empty($this->connections[$key])) {
+                        unset($this->connections[$key]);
+                    }
+                    return;
+                }
```
kelunik: We only push the connection to the array two lines below, so there shouldn't be any entry here to delete.

src/Connection/Http1Connection.php
```diff
@@ -225,14 +258,17 @@ private function doRead(
                             }
                         }
 
-                        if ($this->shouldCloseSocketAfterResponse($response) || $parser->getState() === Http1Parser::BODY_IDENTITY_EOF) {
-                            $this->socket->close();
+                        if ($timeout = $this->determineKeepAliveTimeout($response)) {
```
kelunik: We need to keep the `$parser->getState() === Http1Parser::BODY_IDENTITY_EOF` check here.

src/Connection/Http1Connection.php
```diff
@@ -332,26 +368,44 @@ private function normalizeTraceRequest(Request $request): Request
...
-            return true;
+        if (!\strcasecmp($requestConnHeader, 'close')) {
+            return 0;
+        }
+
+        if ($response->getProtocolVersion() === '1.0') {
+            return 0;
+        }
+
+        if (\strcasecmp($responseConnHeader, 'keep-alive')) {
```
kelunik: Should be checking against `close`, too, as `keep-alive` is the default for HTTP/1.1.

src/Connection/Http1Connection.php
```diff
@@ -332,26 +368,44 @@ private function normalizeTraceRequest(Request $request): Request
...
         }
 
-        if (!$responseConnHeader && $response->getProtocolVersion() === '1.0') {
-            return true;
+        if ($this->requestCounter >= $params['max'] ?? \PHP_INT_MAX) {
+            return 0;
         }
 
-        return false;
+        return \min(\max(0, $params['timeout'] ?? 0), self::MAX_KEEP_ALIVE_TIMEOUT);
```
kelunik: Why is `0` the default value here?

trowski: It's only `0` if the timeout value is negative or unspecified. I'm not sure if that's correct, can a `keep-alive` header have only a `max` value but no `timeout` value?

src/Connection/Http1Connection.php
trowski: If the value is anything other than `'keep-alive'` the connection will be closed.

src/Connection/Http1Connection.php
kelunik: `max` can be ignored if no pipelining is used: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Keep-Alive

trowski: Cool, that simplifies things a bit.

src/Connection/DefaultConnectionPool.php
trowski: The promise is pushed to the array when it's created, so that needs to be deleted.

src/Connection/DefaultConnectionPool.php
```diff
@@ -36,73 +36,106 @@ public function __construct(?Connector $connector = null)
...
-                $connectContext = $connectContext->withTlsContext($tlsContext);
+                ++$index;
+            } else {
+                $this->connections[$key] = [];
+                $index = 0;
             }
 
-            try {
-                $checkoutCancellationToken = new CombinedCancellationToken($cancellation, new TimeoutCancellationToken($request->getTcpConnectTimeout()));
+            $promise = $this->connections[$key][$index] = call(function () use ($request, $isHttps, $authority, $cancellation, $key, $index) {
```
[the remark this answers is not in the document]
trowski: To use `$index` when the promise or connection is removed.

src/Connection/DefaultConnectionPool.php
```diff
@@ -132,6 +133,7 @@ public function getConnection(Request $request, CancellationToken $cancellation)
                     return;
                 }
 
+                // Replace Promise with Connection object.
```
kelunik: Why not just keep the `Promise` and always do `$connection = yield $connection`?

trowski: That would be simpler. Will change.

src/Connection/Http2Connection.php
```diff
@@ -75,6 +75,9 @@ final class Http2Connection implements Connection
     public const INADEQUATE_SECURITY = 0xc;
     public const HTTP_1_1_REQUIRED = 0xd;
 
+    public const DEFAULT_MAX_HEADER_SIZE = 1 << 20;
```
kelunik: 1MB for headers is a bit much, no?

trowski: Probably. We'll have to look at if and how we want these settings to be configured and pick some defaults.

bwoebi: Should not matter too much as a client. TBH, if you do random requests to not-somewhat-known sites, then you need sanity checks at every layer.

src/Connection/Connection.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    public const MAX_KEEP_ALIVE_TIMEOUT = 60;
+
+    /**
+     * @param Request                $request
+     * @param CancellationToken|null $token
+     *
+     * @return Promise<Response>
+     */
+    public function request(Request $request, ?CancellationToken $token = null): Promise;
```
kelunik: `CancellationToken` isn't optional in `ConnectionPool`, so probably can be required here, too?

trowski: Yeah, the `Client` object should take care of creating one.

src/Connection/DefaultConnectionPool.php
```diff
@@ -0,0 +1,144 @@
...
+                foreach ($this->connections[$key] as $index => $connection) {
+                    $connection = yield $connection;
+                    \assert($connection instanceof Connection);
+
+                    if (!$connection->isBusy()) {
+                        return $connection;
+                    }
+                }
+
+                ++$index;
```
kelunik: This isn't safe. Suppose we have 3 concurrent connections which all allow just one request. Due to the `yielding` above, the second and third both have `$index = 0` and hang in the `yield` until the connection is established. Both push to `$index = 1` then.

trowski: Ah, right, since the new connections won't be added to the list used by foreach. Hmm… might have to use a queue or something where that would be the case.

kelunik: I'd just do a keyless push and query the last key.

src/Response.php
```diff
@@ -222,14 +222,6 @@ public function getBody(): Payload
         return $this->body;
     }
 
-    public function withPreviousResponse(?Response $previousResponse): self
-    {
-        $clone = clone $this;
-        $clone->previousResponse = $previousResponse;
-
-        return $clone;
-    }
```
kelunik: Why did you remove this?

trowski: The method was duplicated somehow when I was rebasing.

src/Connection/Http2Connection.php
```diff
@@ -308,8 +314,12 @@ private function run(): \Generator
                     \assert($promise === null || $promise instanceof Promise);
                 }
             }
+
+            var_dump($chunk);
```
kelunik: Leftover.

src/SocketClient.php
```diff
@@ -27,14 +20,14 @@ final class SocketClient implements Client
...
     public const DEFAULT_USER_AGENT = 'Mozilla/5.0 (compatible; amphp/http-client)';
 
-    private $socketPool;
-    private $driverFactory;
+    private $connector;
+    private $connectionPool;
     private $networkInterceptors;
 
-    public function __construct(?SocketPool $socketPool = null, ?Driver\HttpDriverFactory $driverFactory = null)
+    public function __construct(?Connector $connector = null, ?Connection\ConnectionPool $driverFactory = null)
```
kelunik: Should be named $connectionPool.

PR #197, opened 2019-10-01
Automatically retry idempotent requests

kelunik:
This commit adds automatic retries for idempotent methods and favors new connections for non-idempotent requests unless an explicit timeout is available.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Connection_management_in_HTTP_1.x

I'm not sure whether this should really be in `Client`, but I'd like to have it enabled by default.

### DIFF
--- a/src/Connection/Http2Connection.php
+++ b/src/Connection/Http2Connection.php
@@ -243,6 +243,11 @@ final class Http2Connection implements Connection
         return $this->remainingStreams <= 0 || $this->socket->isClosed();
     }
 
+    public function getRemainingTime(): ?int
+    {
+        return null; // TODO Proper implementation
+    }
+
     public function onClose(callable $onClose): void
     {
         if ($this->onClose === null) {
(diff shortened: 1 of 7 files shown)


src/Connection/Http2Connection.php
```diff
@@ -243,6 +243,11 @@ public function isBusy(): bool
         return $this->remainingStreams <= 0 || $this->socket->isClosed();
     }
 
+    public function getRemainingTime(): ?int
```
trowski: I don't believe there is sensible way to implement this with HTTP/2, so I do not think it should be part of the `Connection` interface.

trowski:
We could add a `UnprocessedHttpException` that indicates that a request failed and can safely be retried. The client could automatically retry those requests. The HTTP/2 driver could throw this for any request where the stream was not processed, or it could be thrown for any idempotent request (as defined by the driver).

kelunik:
Closing in favor of #198.

PR #198, opened 2019-10-04
Retry requests when possible

trowski:
Different approach compared to #197.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -7,6 +7,7 @@ use Amp\Http\Client\Connection\ConnectionPool;
 use Amp\Http\Client\Connection\DefaultConnectionPool;
 use Amp\Http\Client\Connection\Stream;
 use Amp\Http\Client\Interceptor\DecompressResponse;
+use Amp\Http\Client\Interceptor\FollowRedirects;
 use Amp\Http\Client\Interceptor\RetryRequests;
 use Amp\Http\Client\Interceptor\SetRequestHeaderIfUnset;
 use Amp\Http\Client\Internal\InterceptedStream;
@@ -24,24 +25,26 @@ final class Client
     private $connectionPool;
 
     /** @var ApplicationInterceptor[] */
-    private $applicationInterceptors;
-    /** @var ApplicationInterceptor[] */
-    private $defaultApplicationInterceptors;
+    private $applicationInterceptors = [];
+
     /** @var NetworkInterceptor[] */
     private $networkInterceptors = [];
 
     /** @var NetworkInterceptor[] */
     private $defaultNetworkInterceptors;
 
-    public function __construct(?ConnectionPool $connectionPool = null, int $retryLimit = 3)
+    /** @var RetryRequests */
+    private $retryInterceptor;
+
+    /** @var FollowRedirects */
+    private $followRedirects;
+
+    public function __construct(?ConnectionPool $connectionPool = null)
     {
         $this->connectionPool = $connectionPool ?? new DefaultConnectionPool;
 
         $this->followRedirects = new FollowRedirects;
-
-        $this->defaultApplicationInterceptors = [
-            new RetryRequests($retryLimit),
-        ];
+        $this->retryInterceptor = new RetryRequests(3);
 
         // We want to set these by default if the user doesn't choose otherwise
         $this->defaultNetworkInterceptors = [
@@ -69,17 +72,8 @@ final class Client
             $request = new Request($requestOrUri);
         }
 
-        $client = clone $this;
-
-        if ($client->defaultApplicationInterceptors) {
-            $client->applicationInterceptors = \array_merge(
-                $client->applicationInterceptors,
-                $client->defaultApplicationInterceptors
-            );
-            $client->defaultApplicationInterceptors = [];
-        }
-
-        if ($client->applicationInterceptors) {
+        if ($this->applicationInterceptors) {
+            $client = clone $this;
             $interceptor = \array_shift($client->applicationInterceptors);
             return $interceptor->request($request, $cancellation, $client);
         }
@@ -90,12 +84,18 @@ final class Client
             return $this->followRedirects->request($request, $cancellation, $client);
         }
 
-        return call(function () use ($client, $request, $cancellation) {
-            $stream = yield $client->connectionPool->getStream($request, $cancellation);
+        if ($this->retryInterceptor) {
+            $client = clone $this;
+            $client->retryInterceptor = null;
+            return $this->retryInterceptor->request($request, $cancellation, $client);
+        }
+
+        return call(function () use ($request, $cancellation) {
+            $stream = yield $this->connectionPool->getStream($request, $cancellation);
 
             \assert($stream instanceof Stream);
 
-            $networkInterceptors = \array_merge($client->defaultNetworkInterceptors, $client->networkInterceptors);
+            $networkInterceptors = \array_merge($this->defaultNetworkInterceptors, $this->networkInterceptors);
             $stream = new InterceptedStream($stream, ...$networkInterceptors);
 
             return yield $stream->request($request, $cancellation);
@@ -141,6 +141,16 @@ final class Client
         return $client;
     }
 
+    /**
+     * @param int $retryLimit Maximum number of times a request may be retried. Only certain requests will be retried
+     *                        automatically (GET, HEAD, PUT, and DELETE requests are automatically retried, or any
+     *                        request that was indicated as unprocessed by the HTTP server).
+     */
+    public function setRetryLimit(int $retryLimit): void
+    {
+        $this->retryInterceptor = new RetryRequests($retryLimit);
+    }
+    
     /**
      * Returns a client that will automatically request the URI supplied by a redirect response (3xx status codes)
      * and return that response instead of the redirect response.
--- a/src/Client.php
+++ b/src/Client.php
@@ -148,6 +148,11 @@ final class Client
      */
     public function setRetryLimit(int $retryLimit): void
     {
+        if ($retryLimit <= 0) {
+            $this->retryInterceptor = null;
+            return;
+        }
+
         $this->retryInterceptor = new RetryRequests($retryLimit);
     }
     
--- a/src/Client.php
+++ b/src/Client.php
@@ -6,6 +6,7 @@ use Amp\CancellationToken;
 use Amp\Http\Client\Connection\ConnectionPool;
 use Amp\Http\Client\Connection\DefaultConnectionPool;
 use Amp\Http\Client\Connection\Stream;
+use Amp\Http\Client\Connection\UnprocessedRequestException;
 use Amp\Http\Client\Interceptor\DecompressResponse;
 use Amp\Http\Client\Interceptor\FollowRedirects;
 use Amp\Http\Client\Interceptor\RetryRequests;
@@ -33,10 +34,10 @@ final class Client
     /** @var NetworkInterceptor[] */
     private $defaultNetworkInterceptors;
 
-    /** @var RetryRequests */
+    /** @var RetryRequests|null */
     private $retryInterceptor;
 
-    /** @var FollowRedirects */
+    /** @var FollowRedirects|null */
     private $followRedirectsInterceptor;
 
     public function __construct(?ConnectionPool $connectionPool = null)
@@ -98,7 +99,11 @@ final class Client
             $networkInterceptors = \array_merge($this->defaultNetworkInterceptors, $this->networkInterceptors);
             $stream = new InterceptedStream($stream, ...$networkInterceptors);
 
-            return yield $stream->request($request, $cancellation);
+            try {
+                return yield $stream->request($request, $cancellation);
+            } catch (UnprocessedRequestException $exception) {
+                throw $exception->getPrevious();
+            }
         });
     }
 
--- a/src/Client.php
+++ b/src/Client.php
@@ -7,7 +7,7 @@ use Amp\Http\Client\Connection\ConnectionPool;
 use Amp\Http\Client\Connection\DefaultConnectionPool;
 use Amp\Http\Client\Connection\Stream;
 use Amp\Http\Client\Interceptor\DecompressResponse;
-use Amp\Http\Client\Interceptor\FollowRedirects;
+use Amp\Http\Client\Interceptor\RetryRequests;
 use Amp\Http\Client\Interceptor\SetRequestHeaderIfUnset;
 use Amp\Http\Client\Internal\InterceptedStream;
 use Amp\NullCancellationToken;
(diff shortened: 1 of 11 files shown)


trowski:
@kelunik The code to retry is in an interceptor, but it could also go directly into `Client` as in your PR. The main difference here is the addition of an exception indicating that the request can be retried, regardless of the request method.

src/Client.php
```diff
@@ -23,22 +24,28 @@ final class Client
...
     private $applicationInterceptors;
+    /** @var ApplicationInterceptor[] */
+    private $defaultApplicationInterceptors;
     /** @var NetworkInterceptor[] */
     private $networkInterceptors;
     /** @var NetworkInterceptor[] */
     private $defaultNetworkInterceptors;
 
-    public function __construct(?ConnectionPool $connectionPool = null)
+    public function __construct(?ConnectionPool $connectionPool = null, int $retryLimit = 3)
```
kelunik: I prefer having a setter here and no constructor argument.

src/Connection/Http1Connection.php
```diff
@@ -449,10 +453,9 @@ private function determineKeepAliveTimeout(Response $response): int
 
         $params = Http\createFieldValueComponentMap(Http\parseFieldValueComponents($response, 'keep-alive'));
 
-        return $this->priorTimeout = \min(
-            \max(0, ((int) ($params['timeout'] ?? $this->priorTimeout)) - 1),
-            self::MAX_KEEP_ALIVE_TIMEOUT
-        );
+        $timeout = (int) ($params['timeout'] ?? $this->priorTimeout);
+
+        return $this->priorTimeout = \min(\max(0, $timeout - 1), self::MAX_KEEP_ALIVE_TIMEOUT);
```
kelunik: I guess the timeout here should be the real timeout. Instead we should avoid trying non-retry-able requests if we're close to the timeout.

src/Connection/UnprocessedRequestException.php
```diff
@@ -0,0 +1,23 @@
...
+    public function __construct(Request $request, HttpException $previous)
+    {
+        parent::__construct("The request was not processed and can be safely retried", 0, $previous);
+        $this->request = $request;
+    }
+
+    public function getRequest(): Request
+    {
+        return clone $this->request;
+    }
```
kelunik: Maybe we should move this method and `isRetryable()` into `HttpException`?

src/Interceptor/RetryRequests.php
```diff
@@ -0,0 +1,55 @@
...
+use Amp\Http\Client\Request;
+use Amp\Promise;
+use function Amp\call;
+
+final class RetryRequests implements ApplicationInterceptor
+{
+    /** @var int */
+    private $retryLimit;
+
+    public function __construct(int $retryLimit = 2)
```
kelunik: The limit here should probably match the one in `Client`, maybe skip providing a default value here.

src/InvalidRequestException.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Amp\Http\Client;
+
+final class InvalidRequestException extends \Exception
```
kelunik: This should still extend the base exception, so `HttpException`.

trowski: I didn't extend `HttpException` so that `RetryRequests` would not catch the exception, but instead I can catch and immediately re-throw.

src/Connection/UnprocessedRequestException.php
trowski: Not everything extending `HttpException` necessarily has a `Request` object to associate with it. Maybe another exception extending `HttpException` that I can use as a base between the two?

src/Connection/Http1Connection.php
trowski: Sure, but then why not just close the connection because it won't get used again anyway?

src/Interceptor/RetryRequests.php
trowski: Oops, not matching was a typo. Required in the interceptor makes sense.

src/Connection/Http1Connection.php
kelunik: A `GET` request coming shortly after might still use that connection, because it can be safely retried if it fails.

trowski: Yes, I missed "non-retry-able."

Pushed an update similar to what you did then.

src/Client.php
```diff
@@ -110,4 +123,14 @@ public function addApplicationInterceptor(ApplicationInterceptor $applicationInt
...
     }
+
+    /**
+     * @param int $retryLimit Maximum number of times a request may be retried. Only certain requests will be retried
+     *                        automatically (GET, HEAD, PUT, and DELETE requests are automatically retried, or any
+     *                        request that was indicated as unprocessed by the HTTP server).
+     */
+    public function setRetryLimit(int $retryLimit): void
+    {
+        $this->retryInterceptor = new RetryRequests($retryLimit);
```
kelunik: Passing `0` should probably directly set `$this-retryInterceptor = null`.

src/Connection/Http2Connection.php
```diff
@@ -1444,8 +1449,13 @@ private function shutdown(?int $lastId = null, ?\Throwable $reason = null): Prom
         $this->socket->close();
 
         if (!empty($this->streams)) {
-            $exception = new SocketException("Connection closed", $reason->getCode(), $reason);
+            $reason = $exception = $reason ?? new SocketException("Connection closed");
             foreach ($this->streams as $id => $stream) {
+                $exception = $reason;
+                if ($id > $lastId && $stream->request !== null) {
```
kelunik: I'm not sure why `$lastId` is defined the way it is. It seems like in `$lastId = $lastId ?? ($id ?? 0);` `$id` is always undefined and why does it fall back to `0`? While it may make sense for the written `GOAWAY` frame, these requests shouldn't count as unprocessed.

src/Interceptor/RetryRequests.php
```diff
@@ -0,0 +1,53 @@
...
+
+    public function __construct(int $retryLimit)
+    {
+        $this->retryLimit = $retryLimit;
+    }
+
+    public function request(Request $request, CancellationToken $cancellation, Client $client): Promise
+    {
+        return call(function () use ($request, $cancellation, $client) {
+            $attempts = 0;
```
kelunik: I think we should start counting at `$attempt = 1`.

src/Request.php
```diff
@@ -296,6 +296,12 @@ public function setBodySizeLimit(int $bodySizeLimit): void
         $this->bodySizeLimit = $bodySizeLimit;
     }
 
+    public function isRetryable(): bool
```
kelunik: Should we rename this to `isIdempotent()`?

src/RequestException.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Amp\Http\Client;
+
+abstract class RequestException extends HttpException
```
kelunik: Seems like this extra exception is useless?

src/Connection/Http2Connection.php
```diff
@@ -309,6 +310,12 @@ private function request(Request $request, CancellationToken $token): Promise
         return call(function () use ($id, $request, $token): \Generator {
             $this->pendingRequests[$id] = $deferred = new Deferred;
 
+            if ($this->socket->isClosed()) {
+                throw new SocketException('Socket closed before request was made');
+            }
+
+            $this->socket->reference();
+
```
kelunik: I've applied this to `master`, because a test was failing on Travis with this error.

https://github.com/amphp/http-client/commit/31904f8f83f1cd8d9bf75adb17ea6ea538880b1e

src/Connection/Http2Connection.php
```diff
@@ -309,6 +310,12 @@ private function request(Request $request, CancellationToken $token): Promise
         return call(function () use ($id, $request, $token): \Generator {
             $this->pendingRequests[$id] = $deferred = new Deferred;
 
+            if ($this->socket->isClosed()) {
+                throw new SocketException('Socket closed before request was made');
```
kelunik: In case the socket closed before we sent anything, we can also safely retry.

src/Connection/Http2Connection.php
trowski: Copy/paste failure from `http-server`. `$id` there is set by a foreach loop that I removed. Will fix.

src/RequestException.php
trowski: Sort of… it acts as shared code for `InvalidRequestException` and `UnprocessedRequestException`. Certainly open to different suggestions.

src/Connection/Http1Connection.php
kelunik: In an optimal world that time would be based on the RTT to the server, it probably makes sense to make it configurable.

src/RequestException.php
kelunik: Where is `::getRequest()` used? Or is it only intended for consumers of the library?

trowski: I originally intended to use it in `RetryRequests` when an `UnprocessedRequestException` was thrown. A consumer *might* use it to see if the request was modified by interceptors, but I'm not sure if it's necessary.

trowski:
Rebased on master.

src/Connection/DefaultConnectionPool.php
```diff
@@ -82,121 +96,27 @@ public function getStream(Request $request, CancellationToken $cancellation): Pr
...
-                });
+                return $connection->getStream($request);
+            }
 
-                return $connection;
-            });
+            $promise = new Coroutine($this->createConnection($request, $cancellation, $authority, $isHttps));
 
-            $this->connections[$key] = $this->connections[$key] ?? new \SplObjectStorage;
+            $this->connections[$key] = $this->connections[$key] ?? $connections;
```
kelunik: I think the `new \SplObjectStorage` was better here.

src/Connection/Http2Connection.php
```diff
@@ -1444,31 +1448,40 @@ private function shutdown(?int $lastId = null, ?\Throwable $reason = null): Prom
...
 
-        Loop::cancel($this->pingWatcher);
+            Loop::cancel($this->pingWatcher);
 
-        return $promise;
+            $bytes = yield $promise;
+
+            $this->socket->close();
+
+            return $bytes;
```
kelunik: Shouldn't this be a `Promise<void>` return?

src/Connection/DefaultConnectionPool.php
trowski: `SplObjectStorage` and foreach didn't work the way I thought, so I pushed an alternative version.

src/Connection/DefaultConnectionPool.php
```diff
@@ -51,27 +55,47 @@ public function getStream(Request $request, CancellationToken $cancellation): Pr
...
+            while ($hashes = \array_keys($this->connections[$key] ?? [])) {
+                if (isset($hash) && $hash !== $hashes[$i++]) {
+                    $i = 0; // Connection list was modified while waiting, start over from beginning.
+                }
+
+                if (!isset($hashes[$i])) {
+                    break;
+                }
+
+                $hash = $hashes[$i];
```
kelunik: This is very hard to read now.

src/Interceptor/RetryRequests.php
```diff
@@ -0,0 +1,64 @@
...
+                    }
+                    // Request can safely be retried.
+                    $exceptions[] = $exception;
+                }
+            } while ($attempts <= $this->retryLimit);
+
+            throw new HttpException(
+                'Request failed after ' . $attempts . ' attempts',
+                0,
+                new MultiReasonException($exceptions)
```
kelunik: I think we should just throw the last exception here. I have the feeling people often don't understand `MultiReasonException` correctly and you need specialized handling when logging to find the actual cause.

test/TimeoutTest.php
```diff
@@ -26,6 +26,7 @@ public function setUp(): void
         parent::setUp();
 
         $this->client = new Client;
+        $this->client->setRetryLimit(0);
```
kelunik: I think we shouldn't retry on timeouts, so this test should be unaffected.

trowski: Really? Request timeouts are the thing I see most often in my logs for request failures.

src/Interceptor/RetryRequests.php
trowski: `MultiReasonException` has certainly been a point of confusion. I dislike hiding information, but 99% of the time if a request fails after multiple attempts, it's probably for the same reason. I'll revert this commit.

src/Connection/DefaultConnectionPool.php
trowski: The only thing I can immediately think to do is move the increment on `$i`. Suggestions?

test/TimeoutTest.php
kelunik: I think we should use the configured timeouts and not multiply these by X by default. The reason for including a retry by default is mostly due to connections closing unexpectedly.

src/Connection/DefaultConnectionPool.php
kelunik: Do we really care about modifications while iterating here? The only case where we really wait is having one connection and that one is pending.

trowski: True, now that we only wait with one connection there isn't a chance for modifications while waiting. That will make this simple.

trowski:
@kelunik I think this is now ready to be merged.

trowski:
Rebased on master again.

src/Client.php
```diff
@@ -32,14 +33,18 @@ final class Client
     /** @var NetworkInterceptor[] */
     private $defaultNetworkInterceptors;
 
-    /** @var FollowRedirects|null */
-    private $followRedirects;
+    /** @var RetryRequests */
```
kelunik: Nullability is missing now.

src/Connection/DefaultConnectionPool.php
```diff
@@ -82,141 +95,171 @@ public function getStream(Request $request, CancellationToken $cancellation): Pr
...
                 \assert($connection instanceof Connection);
             } catch (\Throwable $exception) {
                 // Connection failed, remove from list of connections.
-                $this->connections[$key]->detach($promise);
+                unset($this->connections[$key][$hash]);
 
-                if (!$this->connections[$key]->count()) {
+                if (empty($this->connections[$key])) {
                     unset($this->connections[$key]);
                 }
```
kelunik: This could be extracted into a `dropConnection` method, as it's duplicated below.

src/Connection/DefaultConnectionPool.php
```diff
@@ -82,141 +95,171 @@ public function getStream(Request $request, CancellationToken $cancellation): Pr
...
+
+        return new Http1Connection($socket);
+    }
+
+    /**
+     * @param int $timeout Number of milliseconds before the estimated connection timeout that a non-idempotent
+     *                     request should will not be sent on an existing HTTP/1.x connection, instead opening a
+     *                     new connection for the request. Default is 2000 ms.
+     */
+    public function setConnectionTimeoutDelta(int $timeout): void
```
kelunik: `setTimeoutGracePeriod`?

trowski: Yes, much better. Should the pool be immutable, or no?

src/Interceptor/RetryRequests.php
```diff
@@ -0,0 +1,55 @@
...
+                    if (!$request->isIdempotent()) {
+                        throw $exception;
+                    }
+                    // Request can safely be retried.
+                }
+            } while ($attempts++ <= $this->retryLimit);
+
+            if ($exception instanceof UnprocessedRequestException) {
+                throw $exception->getPrevious();
+            }
```
kelunik: If the interceptor is disabled, we should probably add another interceptor that catches `UnprocessedRequestException`s and retries those or at least unwraps the exception there.

src/Connection/DefaultConnectionPool.php
kelunik: If so, we'd have to empty `$connections` on `__clone`.

kelunik:
:tada:

PR #199, opened 2019-10-05
Immutable client

trowski:
Extension of #198. @kelunik thoughts on the last commit?

### DIFF
--- a/docs/interceptors.md
+++ b/docs/interceptors.md
@@ -12,7 +12,7 @@ use Amp\Http\Client\Client;
 use Amp\Http\Client\Interceptor\SetRequestHeader;
 
 $client = new Client;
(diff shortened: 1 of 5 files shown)


enumag:
:+1: Yeah, immutable client would be really nice. I could then have it as a service in my dependency injection container and add additional interceptors as needed without touching the service itself. Ofc I can still do it with cloning but this feels better.

kelunik:
@trowski Could you modify this PR to only contain the immutability change, please? I'd like to get this merged, as #200 will probably depend on this.

trowski:
@kelunik So remove the follow redirects portion? I think that should be done as well, but I can put that in another PR after this is merged. I can rebase after #198 is merged, since that has diverged a bit now too.

trowski:
Rebasing on master was simple, so I did that quickly. Will need to rebase #198 on this after merging.

Issue #195, opened 2019-09-29
Implement Request Reliability Mechanisms in HTTP/2


kelunik:
See https://http2.github.io/http2-spec/#rfc.section.8.1.4.

trowski:
No, but it should be simple. I'll do another quick PR.

kelunik:
This has been implemented via #202 and #198.

PR #202, opened 2019-10-08
Mark streams refused by server as unprocessed

trowski:
Along with #198, this should fully implement #195.


trowski:
Rebased on master and marked a few other conditions as unprocessed.

PR #203, opened 2019-10-08
Drop RequestException, use exception whitelist for retry

kelunik:


src/Interceptor/RetryRequests.php
```diff
@@ -27,23 +24,21 @@ public function __construct(int $retryLimit)
...
             do {
                 try {
                     return yield $client->request(clone $request, $cancellation);
                 } catch (UnprocessedRequestException $exception) {
                     // Request was deemed retryable by connection, so carry on.
-                } catch (InvalidRequestException | ParseException | TimeoutException $exception) {
-                    // Request is or response is invalid or request timed out, so do not retry.
-                    throw $exception;
-                } catch (HttpException $exception) {
+                } catch (SocketException $exception) {
```
trowski: For this to work, I think `Http2Exception` needs to extend `SocketException`.

kelunik: I've looked through the list of exceptions and most of them seem to be protocol violations, no?

trowski: Yes. I suppose if the server violates the protocol, retrying may not help.

PR #204, opened 2019-10-08
Forbid serialization for most classes

kelunik:
Maybe we should even make this a trait?

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -55,6 +55,11 @@ final class Client
         ];
     }
 
+    public function __sleep(): array
+    {
+        throw new \Error('Serialization of class ' . __CLASS__ . ' is not allowed');
+    }
+
     /**
      * Asynchronously request an HTTP resource.
      *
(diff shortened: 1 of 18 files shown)


trowski:
Is this really necessary? We probably should prevent cloning of a lot of classes too, but we don't currently prevent that either.

trowski:
Another thought: Are there libraries that assume that a public `__sleep` method means the class *is* serializable? Another approach is declaring the method private.

kelunik:
I think it doesn't hurt to forbid it and might prevent issues because people accidentally do dump things, e.g. passing things to parallel that shouldn't be serialized.

I've taken the approach from https://ocramius.github.io/extremely-defensive-php/#/, might be better to simply make the method private to have it better detectable for static analysis.

trowski:
Should we add a couple traits like `ForbidSerialization` and `ForbidCloning` to `amphp/amp` so we can reuse them everywhere?

kelunik:
@trowski Seems fine with me. Can traits have final methods?

trowski:
Good question, [seems they can](https://3v4l.org/Clhck).

enumag:
Seems unnecessary to me.

PR #210, opened 2019-10-10
Move liveliness check to Connection interface

trowski:
This is an attempt to add a method to the `Connection` interface that can be used test a connection to determine if it should be used for a request, particular a non-idempotent request.

Not sure on naming or the approach for HTTP/1.x connections.

### DIFF
--- a/src/Connection/Connection.php
+++ b/src/Connection/Connection.php
@@ -13,31 +13,24 @@ interface Connection
     public const MAX_KEEP_ALIVE_TIMEOUT = 60;
 
     /**
+     * @return Promise<Stream>
+     */
+    public function getStream(): Promise;
+
+    /**
      * @param Request $request
      *
-     * @return Stream
+     * @return Promise<Stream|null> Returns a stream for the given request, or null if no stream is available or if
+     *                              the connection is not suited for the given request.
      *
      * @throws SocketException If all available streams have been used. Verify a stream is available with isBusy().
      */
-    public function getStream(Request $request): Stream;
+    public function getStreamFor(Request $request): Promise;
 
     /**
      * @return string[] Array of supported protocol versions.
      */
     public function getProtocolVersions(): array;
-
-    /**
-     * @return bool True if a stream is still available, false if the connection is completely busy.
-     */
-    public function hasStreamAvailable(): bool;
-
-    /**
-     * @param Request $request
-     *
-     * @return Promise<bool> True if the connection is safe to use for a new request, false if a new connection should
-     *                       be opened.
-     */
-    public function checkLiveliness(Request $request): Promise;
 
     public function close(): Promise;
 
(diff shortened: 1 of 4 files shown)


src/Connection/DefaultConnectionPool.php
```diff
@@ -100,23 +100,16 @@ public function getStream(Request $request, CancellationToken $cancellation): Pr
...
-                if ($connection instanceof Http2Connection && $connection->isIdle() && !(yield $connection->ping())) {
-                    continue; // Connection closed while idle.
-                }
-
-                if ($connection instanceof Http1Connection
-                    && $connection->getRemainingTime() < $this->timeoutGracePeriod
-                    && !$request->isIdempotent()
-                ) {
-                    continue; // Connection is at high-risk of closing before the request can be sent.
+                if (!yield $connection->checkLiveliness()) {
```
trowski: Should this check be limited to non-idempotent requests?

kelunik: Maybe just pass the request in and let the connection decide?

trowski: Yeah, that would work. `checkLiveliness` is maybe the wrong name? Have a succinct suggestion instead of `isConnectionSuitedForRequest`?

kelunik: isReuseAllowed?

src/Connection/Connection.php
```diff
@@ -12,25 +12,26 @@ interface Connection
 {
     public const MAX_KEEP_ALIVE_TIMEOUT = 60;
 
+    /**
+     * @return Promise<Stream>
+     */
+    public function getStream(): Promise;
```
kelunik: Do we really need this separate method?

trowski: The purpose was that the initial request for a stream from a connection should be independent of the request and never resolve with null.

Is that really necessary? Probably not, the connection can know if it's the first time `getStreamFor()` is called and never fail.

src/Connection/Http2Connection.php
```diff
@@ -206,29 +206,35 @@ public function getProtocolVersions(): array
...
     {
         if (!$this->initialized || $this->settingsDeferred !== null) {
             throw new \Error('The promise returned from ' . __CLASS__ . '::initialize() must resolve before using the connection');
         }
 
-        if ($this->remainingStreams <= 0) {
-            throw new \Error('All available streams have been used');
-        }
+        return call(function () use ($request) {
+            if ($this->streamId > 0 && empty($this->streams) && !yield $this->ping()) {
```
kelunik: This will ping multiple times concurrently if the condition is met.

trowski: `ping()` only allows one un-acked ping to be in-flight at a time. Successive calls return the same promise.

PR #211, opened 2019-10-14
Forbid user info component in URIs

trowski:
Alternative to #209, forbidding the user info component in request URIs.

The placement of the check in `Client` is to allow an application interceptor to set an *Authorization* header if desired. Otherwise if we want to totally disallow it, the check could be moved higher.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -3,6 +3,7 @@
 namespace Amp\Http\Client;
 
 use Amp\CancellationToken;
+use Amp\Failure;
 use Amp\Http\Client\Connection\ConnectionPool;
 use Amp\Http\Client\Connection\DefaultConnectionPool;
 use Amp\Http\Client\Connection\Stream;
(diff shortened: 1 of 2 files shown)


src/Client.php
```diff
@@ -91,6 +91,12 @@ public function request($requestOrUri, ?CancellationToken $cancellation = null):
             return $this->retryInterceptor->request($request, $cancellation, $client);
         }
 
+        if ($request->getUri()->getUserInfo() !== '') {
+            throw new \Error('The user information (username:password) component of URIs has been deprecated '
```
enumag: Not really sure but maybe an `\ErrorException` with `$severity` set to `E_USER_DEPRECATED` might be better?

test/ClientTest.php
```diff
@@ -0,0 +1,26 @@
+<?php /** @noinspection PhpUnhandledExceptionInspection */
```
enumag: Completely unrelated but... any reason why amp isn't using declare strict types?

kelunik: Please don't hijack unrelated issues / PRs like that and open a new issue.

src/Client.php
trowski: I always felt `ErrorException` was meant to be used in user-defined error handlers. Catching the exception isn't really useful either, so that's why I went with `Error`.

kelunik: This should throw an `HttpException`, as the URL might be user input.

PR #215, opened 2019-10-23
Introduce HttpClient interface

kelunik:
This allows decoupling the client entirely from the connection pool and interceptors, which become a implementation detail of specific implementations.

An HttpClientBuilder has been introduced, so users don't need to be aware of the different implementations and how to compose them.

Passing a simple string instead of a Request object has been removed, because it simplifies the public API and because there could be subtle differences now, given that multiple implementations exist.


examples/6-concurrent-requests.php
```diff
@@ -16,7 +16,7 @@
     ];
 
     // Instantiate the HTTP client
-    $client = new Client;
+    $client = HttpClientBuilder::ofPool()->build();
```
kelunik: Maybe we should add another static method `create()` / `createDefault()` / `buildDefault()` which is a shortcut for `::ofPool()->build()`?

trowski: I think so, yes. +1 for `buildDefault()`.

trowski:
`HttpClient` should also include `intercept(NetworkInterceptor $interceptor)`. Right now, when I call `HttpClientBuilder::buildDefault()->intercept(...)`, I get an undefined method warning.

src/HttpClientBuilder.php
```diff
@@ -0,0 +1,121 @@
...
+
+use Amp\Http\Client\Connection\ConnectionPool;
+use Amp\Http\Client\Interceptor\FollowRedirects;
+use Amp\Http\Client\Interceptor\RetryRequests;
+
+final class HttpClientBuilder
+{
+    public static function of(HttpClient $base): self
+    {
+        return new self($base);
```
trowski: I'm concerned this may not work as expected for users. Consider the following:

```php
$client = HttpClientBuilder::buildDefault();
// $client used in DI
$revisedClient = HttpClientBuilder::of($client)
    ->followRedirects(0)
    ->retry(2);
// $revisedClient still will follow redirects and will retry 4 times.

File: src/InterceptedHttpClient.php [hunk 3]
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace Amp\Http\Client;
+
+use Amp\CancellationToken;
+use Amp\NullCancellationToken;
+use Amp\Promise;
+
+final class InterceptedHttpClient implements HttpClient
```
trowski: Should this class be an internal class?

kelunik: I don't see a reason to make it internal.

src/HttpClientBuilder.php
kelunik: What's your suggestion for an improvement here?

src/InterceptedHttpClient.php
trowski: Reasoning was to ensure users use `HttpClientBuilder` instead of this class to add application interceptors.

kelunik:
@trowski I omitted `intercept(...)` on `HttpClient`, because it's now an implementation detail of `PooledHttpClient`, so you can still do:

```php
$httpClient = (new PooledHttpClient)
    ->intercept(...)
    ->intercept(...);

$httpClientBuilder = HttpClientBuilder::of($httpClient);
```

trowski:
So I have to build an `HttpClient` instance and provide it to an object that builds `HttpClient` instances… I'm not thrilled with that API.

kelunik:
Yeah, it's probably not the best API we can deliver, let's discuss some more in IRC.

src/HttpClientBuilder.php
trowski: With the current structure, only to not allow providing a base `HttpClient` instance to the builder.

src/InterceptedHttpClient.php
kelunik: If you want to build another builder, that's fine and you can reuse `InterceptedHttpClient` then.

src/HttpClientBuilder.php
kelunik: I've removed the possibility to pass a custom base, so this is no longer a problem.

PR #218, opened 2019-10-27
Ensure clones of Request are deep clones

### LINKED ISSUE #216: Forbid request cloning
Currently `Request` is still allowed to be cloned, which is used e.g. in `FollowRedirects`, but this will copy the attributes array and its values and share them, resulting in non-sense if things like response timings are stored in the request attributes (see #193).

We could just empty the attributes on cloning, but then we have to remove all other clone operations in the interceptor chain, because those must not null the attributes, otherwise attributes become useless.

I think the best way forward might be to just entirely forbid cloning. Maybe we should introduce an additional property on request that allows to freeze a request object to prevent further modification?

kelunik:
All properties are either immutable for deep clones now.

Fixes #216.

### DIFF
--- a/src/Request.php
+++ b/src/Request.php
@@ -367,27 +367,6 @@ final class Request extends Message
     }
 
     /**
-     * Modify a variable from the request's local storage.
-     *
-     * Each request has its own local storage to which applications and interceptors may read and write data.
-     * Other interceptors which are aware of this data can then access it without the server being tightly coupled to
-     * specific implementations.
-     *
-     * Note: The given callback receives a deep clone of the request's attribute, which can be modified. The result of
-     * the operation will be deeply cloned again, to ensure clones of the request object are deep clones. If the
-     * callback returns null, the passed instance will be used instead of the return value.
-     *
-     * @param string   $name Name of the attribute, should be namespaced with a vendor and package namespace like
-     *     classes.
-     * @param callable $mutator
-     */
-    public function mutateAttribute(string $name, callable $mutator): void
-    {
-        $clone = $this->getAttribute($name);
-        $this->setAttribute($name, $mutator($clone) ?? $clone);
-    }
-
-    /**
      * Assign a variable to the request's local storage.
      *
      * Each request has its own local storage to which applications and interceptors may read and write data.
--- a/src/Request.php
+++ b/src/Request.php
@@ -18,6 +18,16 @@ final class Request extends Message
     public const DEFAULT_HEADER_SIZE_LIMIT = 2 * 8192;
     public const DEFAULT_BODY_SIZE_LIMIT = 10485760;
 
+    private static function clone($value)
+    {
+        if ($value === null || \is_scalar($value)) {
+            return $value;
+        }
+
+        // force deep cloning
+        return \unserialize(\serialize($value), ['allowed_classes' => true]);
+    }
+
     /** @var string[] */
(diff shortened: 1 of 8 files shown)


src/Request.php
```diff
@@ -347,32 +363,56 @@ public function getAttribute(string $name)
...
+     * callback returns null, the passed instance will be used instead of the return value.
+     *
+     * @param string   $name Name of the attribute, should be namespaced with a vendor and package namespace like
+     *     classes.
+     * @param callable $mutator
+     */
+    public function mutateAttribute(string $name, callable $mutator): void
+    {
+        $clone = $this->getAttribute($name);
+        $this->setAttribute($name, $mutator($clone) ?? $clone);
```
kelunik: I'm not sure about this method, yet. It is mostly intended for short closures in PHP 7.4, but usage like `$request->mutateAttribute($name, fn ($attr) => $attr->foo = 'bar');` will result in the attribute becoming `'bar'` instead of `$attr`.

trowski:
The requirement that attributes be serializable makes me rather uncomfortable. This would be a no-go in http-server, so I'm not sure I want to have that kind of restriction in http-client.

kelunik:
Our HTTP server doesn't have any retry mechanism. I thought about it pretty much a whole day and this seems to be the only viable solution to do retries to me.

kelunik:
I thought about that as well, but it doesn't solve the problem with attributes, because we can't ensure they're immutable. We'd have to deeply clone them like in this solution.

This solution has the added benefit that request and response APIs are aligned without response being not really immutable (if we'd change that back, too).

It's basically the best of both worlds, with the only drawback being a slight performance penalty for attributes, but those shouldn't be used too often.

trowski:
Yes – attributes were the reason I didn't like the request being immutable since they weren't really immutable. With this solution, the request could truly be immutable again.

Request headers are still mutable, which I thought was your original concern – `DecompressResponse` sets the `accept-encoding` header, but then on a request retry assumes it did not set the header and does not decode the response.

kelunik:
They're mutable, yes, but as the request object is required to be cloned at every layer, mutability isn't a problem anymore.

trowski:
Requiring cloning is easy to enforce in our code, but I'd be surprised if user interceptors would always remember to clone the request.

kelunik:
That's the exact reason why `HttpClient` instances are required to clone, not interceptors, [I've committed that to master](https://github.com/amphp/http-client/blob/0eaeda0199fafee0bcd9f8af759c90dc54ef210d/src/HttpClient.php#L23-L25) already (by accident).

trowski:
Ok, enforcing that in the client is a much lower bar that I'm more comfortable with. Looks like we need to add `clone` in the `Stream::request()` implementations as well. In turn, we can probably remove `clone` from the `Connection::request()` implementations.

src/Request.php
```diff
@@ -18,6 +18,16 @@ final class Request extends Message
...
     public const DEFAULT_BODY_SIZE_LIMIT = 10485760;
 
+    private static function clone($value)
+    {
+        if ($value === null || \is_scalar($value)) {
+            return $value;
+        }
+
+        // force deep cloning
+        return \unserialize(\serialize($value), ['allowed_classes' => true]);
```
enumag: What about Request::$onPush? Closure is not serializable.

PR #219, opened 2019-11-07
Implement concurrency limits

kelunik:


src/Connection/LimitedConnectionPool.php
```diff
@@ -0,0 +1,62 @@
...
+                    $lock
+                ) {
+                    try {
+                        /** @var Response $response */
+                        $response =  yield $stream->request($request, $cancellationToken);
+                    } finally {
+                        $lock->release();
+                    }
+
+                    // TODO Await body completion
```
kelunik: @trowski We might want to re-introduce the completion promise on `Response` for that that completes as soon as the response is fully read from the socket layer.

PR #220, opened 2019-11-10
Add timing API / HAR support

### LINKED ISSUE #193: Add timing API
- DNS lookup
 - TCP connect
 - TLS handshake
 - Request send
 - TTFB
 - Response receive

kelunik:
Resolves #193.

Tools like http://www.softwareishard.com/har/viewer/ or https://toolbox.googleapps.com/apps/har_analyzer/ can be used for visualization.

### DIFF
--- a/src/Connection/Http1Connection.php
+++ b/src/Connection/Http1Connection.php
@@ -12,6 +12,7 @@ use Amp\Deferred;
 use Amp\Emitter;
 use Amp\Http;
 use Amp\Http\Client\Connection\Internal\Http1Parser;
+use Amp\Http\Client\HarAttributes;
 use Amp\Http\Client\HttpException;
 use Amp\Http\Client\Internal\ForbidCloning;
 use Amp\Http\Client\Internal\ForbidSerialization;
@@ -35,6 +36,7 @@ use Amp\Success;
 use Amp\TimeoutCancellationToken;
 use function Amp\asyncCall;
 use function Amp\call;
+use function Amp\Internal\getCurrentTime;
 
 /**
  * Socket client implementation.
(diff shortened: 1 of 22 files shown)


kelunik:
While looking at the HAR for `php examples/basic/6-customization.php https://example.org/` I noticed strange blocking times. Those are caused by the HTTP/2 pings we do for connection liveness. I guess we should remove these pings, as they cause unnecessary delay.

src/Connection/Http1Connection.php
```diff
@@ -35,6 +36,7 @@
 use Amp\TimeoutCancellationToken;
 use function Amp\asyncCall;
 use function Amp\call;
+use function Amp\Internal\getCurrentTime;
```
kelunik: @trowski We should move `getCurrentTime()` to `Amp\getCurrentTime()` before merging this.

kelunik: Depends on https://github.com/amphp/amp/pull/290.

src/Interceptor/RecordStartTime.php
```diff
@@ -0,0 +1,22 @@
...
+use Amp\Http\Client\Request;
+use Amp\Promise;
+use function Amp\Internal\getCurrentTime;
+
+final class RecordStartTime implements ApplicationInterceptor
+{
+    public function request(Request $request, CancellationToken $cancellation, DelegateHttpClient $next): Promise
+    {
+        $request->setAttribute(HarAttributes::STARTED_DATE_TIME, new \DateTimeImmutable);
+        $request->setAttribute(HarAttributes::TIME_START, getCurrentTime());
```
trowski: Should we just always record these, since many other Har attributes are being recorded in the connection pools.

kelunik: @trowski `HttpClientBuilder` always enables recording these, but they need to be the outer-most interceptor to have the correct time there.

trowski: Ok, so this isn't really an optional interceptor then.

kelunik: It's optional as long as you don't want to write a log and don't use `HttpClientBuilder`.

trowski: Yes, I meant not optional when using `HttpClientBuilder`, I didn't notice it was added there at first.

PR #221, opened 2019-11-11
Set used protocol version to request

kelunik:


trowski:
What's the purpose? `Response::getProtocolVersion()` tells you what protocol version was used.

kelunik:
@trowski With HTTP/1.0, you might get a HTTP/1.1 response, no?

trowski:
Yes, you might receive a 1.1 response, but how does setting it on the `Request` object help?

kelunik:
@trowski It allow printing the request as sent.

PR #223, opened 2019-11-18
Redesign interceptor implementation

### LINKED ISSUE #222: Improve timing API
With #193 we introduced a timing API, but it's rather hard coded and currently doesn't work correctly with redirects and retries.

Maybe we should use an additional event listener system for these things like https://square.github.io/okhttp//events/

kelunik:
~This allows the interceptor chain to be automatically restarted if a new
request is passed.~

New event system allows timing API to be implemented in a nicer an more decoupled way.

Closes #222.


kelunik:
@trowski This might also allow us to drop the cloning requirement and only clone in `HttpClient`?

PR #224, opened 2019-11-21
Support CONNECT over HTTP/2

trowski:


src/Connection/Http2Connection.php
```diff
@@ -1467,9 +1478,58 @@ static function () {
 
                             $this->releaseStream($id); // Response has no body, release stream immediately.
                         } else {
-                            $this->trailerDeferreds[$id] = new Deferred;
                             $this->bodyEmitters[$id] = new Emitter;
 
+                            if ($status >= Status::OK && $status < Status::MULTIPLE_CHOICES && $stream->request->getMethod() === 'CONNECT') {
```
kelunik: I'd use the literal numbers here, as that's more useful for ranges. The names are mostly useful for single status codes, but not ranges.

src/Connection/Http2Connection.php
```diff
@@ -1467,9 +1478,58 @@ static function () {
...
+                                if (($onUpgrade = $stream->request->getUpgradeHandler()) === null) {
+                                    throw new Http2StreamException('CONNECT or upgrade request made without upgrade handler callback', $id, self::CANCEL);
+                                }
+
+                                $stream->bodySizeLimit = \PHP_INT_MAX; // Remove any body size limitation.
```
kelunik: Should we regularly increase it for 32 bit or very-very long running connections?

We probably also have to increase the timeout.

src/Connection/UpgradedStream.php
```diff
@@ -0,0 +1,108 @@
...
+{
+    use ForbidCloning;
+    use ForbidSerialization;
+
+    private $read;
+    private $write;
+    private $localAddress;
+    private $remoteAddress;
+
+    public function __construct(InputStream $read, callable $write, SocketAddress $local, SocketAddress $remote)
```
kelunik: With the `callable` I prefer this class to be internal, I guess.

kelunik: I guess `$write` Should just be an `OutputStream`?

src/Connection/Http2Connection.php
trowski: The idle watcher should return because there will be a stream in use so long as the tunneled connection in open.

src/Connection/UpgradedStream.php
trowski: For a class in the public namespace, yes it should use `OutputStream` instead. Probably the simplest way to do that is wrap the callable in an internal class implementing `OutputStream`.

kelunik:
We could use caddy to test this, I guess.

trowski:
Can we leave the proxy socket hack to `http-tunnel` and return the simple socket here? This package is already getting a bit large.

src/Connection/Http2Connection.php
```diff
@@ -1467,9 +1478,60 @@ static function () {
...
+                                            if ($final) {
+                                                $this->streams[$id]->state |= Http2Stream::LOCAL_CLOSED;
+                                            }
+
+                                            return $this->writeData($data, $id);
+                                        }
+                                    ),
+                                    $this->socket->getLocalAddress(),
+                                    $this->socket->getRemoteAddress()
+                                );
```
kelunik: It's good the user can't access the raw resource here!

src/Connection/Http2Connection.php
```diff
@@ -1467,9 +1478,60 @@ static function () {
...
+
+                                asyncCall(function () use ($onUpgrade, $socket, $stream, $response): \Generator {
+                                    try {
+                                        yield call($onUpgrade, $socket, clone $stream->request, $response);
+                                    } catch (\Throwable $exception) {
+                                        throw new HttpException('Upgrade handler threw an exception', 0, $exception);
+                                    } finally {
+                                        $socket->close();
```
kelunik: It should probably only close here if the upgrade handler throws.

src/Connection/Internal/Http2UpgradeOutputStream.php
```diff
@@ -0,0 +1,45 @@
...
+namespace Amp\Http\Client\Connection\Internal;
+
+use Amp\ByteStream\ClosedException;
+use Amp\ByteStream\OutputStream;
+use Amp\ByteStream\StreamException;
+use Amp\Http\Client\HttpException;
+use Amp\Promise;
+use function Amp\call;
+
+class Http2UpgradeOutputStream implements OutputStream
```
kelunik: This should be `final` and it's missing `/** @internal */`

src/Connection/UpgradedStream.php
```diff
@@ -0,0 +1,99 @@
...
+        if ($this->write === null) {
+            return new Failure(new SocketException('The socket is no longer writable'));
+        }
+
+        return $this->write->end($finalData);
+    }
+
+    public function reference(): void
+    {
+        // No-op
```
kelunik: Should this really be a no-op?

src/Connection/UpgradedStream.php
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    public function unreference(): void
+    {
+        // No-op
+    }
+
+    public function isClosed(): bool
+    {
+        return $this->read === null;
```
kelunik: This will only return true if explicitly closed.

src/Connection/UpgradedStream.php
```diff
@@ -0,0 +1,117 @@
...
+        $this->write = $write;
+        $this->localAddress = $local;
+        $this->remoteAddress = $remote;
+    }
+
+    public function read(): Promise
+    {
+        return call(function (): \Generator {
+            if ($this->read === null) {
+                throw new SocketException('The socket is no longer readable');
```
kelunik: This should return `null`.

kelunik:
It might be better to just throw an exception for `CONNECT` requests on HTTP/2 connections for now and implement this later?

trowski:
@kelunik I'm fine with throwing for now if you want to tag 4.0. Since the implementation got changed around this will have to be re-examined and I don't have the time to do that right now. We can always release a 4.1 with this supported.

PR #231, opened 2019-12-13
Do not set content-length header when upgrading to websocket

danog:


src/Connection/Http1Connection.php
```diff
@@ -563,9 +563,11 @@ private function normalizeRequestBodyHeaders(Request $request): \Generator
         /** @var RequestBody $body */
         $body = $request->getBody();
         $bodyLength = yield $body->getBodyLength();
-
+        
         if ($bodyLength === 0) {
-            $request->setHeader('content-length', '0');
+            if ($request->getHeader('upgrade') !== 'websocket') {
+                $request->setHeader('content-length', '0');
```
trowski: `Request::removeHeader('content-length')` would be better.

kelunik: I guess this should be done for any upgrade request?

danog: Idk @kelunik, I suppose it depends on the spec of the specific protocol, but I'm not sure.
Anyway made the change requested by @trowski

trowski: Yes, the header should be removed for any upgrade request that has a 0-length body.

danog: Ok then

kelunik: I've just fixed this in 626f952e36bb3ca63c7ff9b3ab4e9418a7bbec83 slightly differently, omitting the `content-length` header for any `HEAD` / `GET` / `CONNECT` request. We could also omit it for any bodyless request, but I think we had issues with misbehaving servers in the past.

PR #233, opened 2019-12-17
Avoid problems with league URI port set to null if using default port

danog:


kelunik:
@danog Thanks for catching this. I've applied a slightly different fix, getting rid of the `if`.

danog:
Great!
Do you think you could make a release (or at least tag another RC and update dependency versions for secondary packages?).

Thanks!

kelunik:
I wanted to release today, yes, but @trowski wants to test over the weekend, so I'll tag a new RC.

PR #238, opened 2019-12-31
Add HttpClientBuilder::with/withoutHeaderIfUnset()

nicolas-grekas:
Useful when we want to disable the default interceptors, as in https://github.com/symfony/symfony/pull/35115#pullrequestreview-337184875


nicolas-grekas:
> I think these should be separate functions to opt-out of any of these

updated

kelunik:
I've implemented it slightly differently in #241, what do you think?

nicolas-grekas:
Works for me thanks!

PR #239, opened 2019-12-31
Add Request::onInformationalResponseHandler() to process e.g. 103 Early Hints

nicolas-grekas:
Will be required to complete https://github.com/symfony/symfony/pull/35115
Draft state for now until I confirm this allows passing the `testInformationalResponseStream()` test case.
Still, WDYT?

See https://evertpot.com/http/103-early-hints

### DIFF
--- a/src/Connection/Internal/Http2ConnectionProcessor.php
+++ b/src/Connection/Internal/Http2ConnectionProcessor.php
@@ -2,6 +2,7 @@
 
 namespace Amp\Http\Client\Connection\Internal;
 
+use Amp\ByteStream\InMemoryStream;
 use Amp\ByteStream\IteratorStream;
 use Amp\ByteStream\StreamException;
 use Amp\CancellationToken;
@@ -405,10 +406,6 @@ final class Http2ConnectionProcessor implements Http2Processor
             return;
         }
 
-        if ($status < 200) {
-            return; // ignore 1xx responses
-        }
-
         asyncCall(function () use ($stream, $streamId) {
             try {
                 foreach ($stream->request->getEventListeners() as $eventListener) {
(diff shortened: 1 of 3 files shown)


src/EventListener.php
```diff
@@ -107,6 +107,16 @@ public function completeSendingRequest(Request $request, Stream $stream): Promis
...
 
+    /**
+     * Called by {@see Stream::request()} when a 1xx informational response is received.
+     *
+     * @param Request  $request
+     * @param Response $response
+     *
+     * @return Promise Should resolve successfully, otherwise aborts the request.
+     */
+    public function handleInformationalResponse(Request $request, Response $response): Promise;
```
kelunik: This will be a BC break. My suggestion is to add it as `Request::setInformationalResponseHandler`, like `setPushHandler` and `setUpgradeHandler`. Maybe we can find a slightly shorter name for it.

nicolas-grekas:
> My suggestion is to add it as Request::setInformationalResponseHandler

updated, I confirm this works for Symfony HttpClient

src/Connection/Internal/Http2ConnectionProcessor.php
```diff
@@ -419,6 +416,31 @@ public function handleHeaders(int $streamId, array $pseudo, array $headers): voi
...
+
+            if ($onInformationalResponse !== null) {
+                asyncCall(function () use ($onInformationalResponse, $response, $streamId) {
+                    try {
+                        yield call($onInformationalResponse, $response);
+                    } catch (\Throwable $e) {
+                        $this->handleStreamException(new Http2StreamException('Informational response handler threw an exception', $streamId, self::CANCEL));
+                    }
+                });
+            }
```
kelunik: If we do it async here, we should probably do the same for HTTP/1?

nicolas-grekas: I've no opinion on that: I just borrowed from the way event listener are called. What would you advise?

PR #241, opened 2020-01-01
Allow skipping the default interceptors in HttpClientBuilder

kelunik:
Alternative implementation to #238. Closes #238.


nicolas-grekas:
In the end, I'm not going to use `HttpClientBuilder`, so that this is not needed on Symfony's side anymore.

Issue #243, opened 2020-01-03
How do I cancel an already accepted push promise?


nicolas-grekas:
In my use case, I accept all HTTP/2 push promises and put them in a temporary cache.
After some time, I want to cancel the ones that have not been used yet.
I didn't find a way to do so, did I miss something?
Thanks for the help :)

trowski:
You can throw a `Amp\CancelledException` or an `Amp\Http\Client\HttpException` from the push promise handler to cancel the push promise.

nicolas-grekas:
> You can throw  from the push promise handler

I spotted that, but I'm not sure that answers my use case. Maybe it's me not being fluent enough with amp.
Inside the push handler, I store pushed req+response-promise in a queue.
Then, later on, I consume from the queue. And from time to time, I want to empty the queue.
Could the push handler be give a 3rd argument: a cancellation source that we could use to cancel the pushed req/resp when needed later on?

trowski:
Funny you should suggest a cancellation token source as a third argument, since that's how I originally wrote that function. @kelunik I think we should add that back.

PR #245, opened 2020-01-06
Provide cancellation token source to push handler

trowski:
Related to #243.

### DIFF
--- a/src/Connection/Internal/Http2ConnectionProcessor.php
+++ b/src/Connection/Internal/Http2ConnectionProcessor.php
@@ -591,17 +591,20 @@ final class Http2ConnectionProcessor implements Http2Processor
                 $this->releaseStream($streamId, $exception);
             });
 
+            $promise = $stream->pendingResponse->promise();
+            $promise->onResolve(function () use ($cancellationToken, $cancellationId): void {
+                $cancellationToken->unsubscribe($cancellationId);
+            });
+
             $onPush = $stream->request->getPushHandler();
 
             try {
-                yield call($onPush, $stream->request, $stream->pendingResponse->promise());
+                yield call($onPush, $stream->request, $promise, $tokenSource);
             } catch (HttpException | StreamException | CancelledException $exception) {
                 $tokenSource->cancel($exception);
             } catch (\Throwable $exception) {
                 $tokenSource->cancel($exception);
                 throw $exception;
-            } finally {
-                $cancellationToken->unsubscribe($cancellationId);
             }
         });
     }


src/Connection/Internal/Http2ConnectionProcessor.php
```diff
@@ -594,7 +594,7 @@ static function () {
             $onPush = $stream->request->getPushHandler();
 
             try {
-                yield call($onPush, $stream->request, $stream->pendingResponse->promise());
+                yield call($onPush, $stream->request, $stream->pendingResponse->promise(), $tokenSource);
```
kelunik: Because of the `finally` below it seems like it's only possible to cancel while the promise returned from `$onPush` has not resolved, so it doesn't seem to make anything easier.

trowski: D'oh, yes, that will have to be pushed to the code resolving the promise.

src/Connection/Internal/Http2ConnectionProcessor.php
```diff
@@ -591,17 +591,20 @@ static function () {
                 $this->releaseStream($streamId, $exception);
             });
 
+            $promise = $stream->pendingResponse->promise();
+            $promise->onResolve(function () use ($cancellationToken, $cancellationId): void {
+                $cancellationToken->unsubscribe($cancellationId);
+            });
```
kelunik: This isn't correct either, I guess it has to subscribe to the `$response->getTrailers()` promise.

trowski: Yep. This keeps getting more complicated. 😛

src/Connection/Internal/Http2ConnectionProcessor.php
kelunik: Maybe we should just keep the current behavior?

src/Connection/Internal/Http2ConnectionProcessor.php
kelunik: Maybe we should just keep the current behavior?

trowski: I'm thinking so. The user can return a promise that isn't resolved until they know they want to receive the push.

@nicolas-grekas, you return a deferred promise from the handler, subsequently failing the deferred to cancel the push.

kelunik:
Closing due to https://github.com/amphp/http-client/pull/245#discussion_r365993151.

src/Connection/Internal/Http2ConnectionProcessor.php
nicolas-grekas: Got it, thank you and sorry for the noise, I should have thought about deferred earlier.

PR #250, opened 2020-01-13
Keep original header case

kelunik:
Based on https://github.com/amphp/http/pull/11.


src/Connection/Internal/Http1Parser.php
```diff
@@ -180,7 +180,12 @@ public function parse(string $data = null): ?Response
                 $this->complete = true;
             }
 
-            return new Response($this->protocol, $this->statusCode, $this->statusReason, $this->headers, new InMemoryStream, $this->request);
+            $response = new Response($this->protocol, $this->statusCode, $this->statusReason, [], new InMemoryStream, $this->request);
+            foreach ($this->headers as [$key, $value]) {
+                $response->addHeader($key, $value);
```
trowski: Rather than a function call for every header, instead build an associative array before the `Response` constructor.

kelunik: That won't work if one header is present with two different cases.

trowski: Don't lowercase the header names when making the associative array. I was thinking simply this:

```php
$headers = [];
foreach ($this->headers as [$name, $value]) {
     $headers[$name][] = $value;
}
$response = new Response($this->protocol, $this->statusCode, $this->statusReason, $headers, new InMemoryStream, $this->request);
```

trowski: Never mind, I just saw where I went wrong, since the constructor uses `setHeaders()`. Since that just does a foreach to `setHeader()`, there's not much difference anyway.

kelunik:
@trowski Why two separate releases?

PR #253, opened 2020-01-15
Add connection-limiting pool

### LINKED ISSUE #252: LimitedConnectionPool limits concurrent requests instead of concurrent connections
`LimitedConnectionPool` is not compatible with multiplexing several HTTP/2 requests on a single connection. Instead, it throttles the number of *streams*, which looks quite counter productive :)

trowski:
Not sure on the name, suggestions welcome.

Should we deprecate `LimitedConnectionPool` or rename it  (with an alias of the old name) to `SingleStreamConnectionPool` or leave it as-is?

Closes #252.


src/Connection/FiniteConnectionPool.php
```diff
@@ -0,0 +1,234 @@
...
+    /** @var int */
+    private $totalConnectionAttempts = 0;
+
+    /** @var int */
+    private $totalStreamRequests = 0;
+
+    /** @var int */
+    private $openConnectionCount = 0;
+
+    public function __construct(int $maxConnections, ?ConnectionFactory $connectionFactory = null)
```
nicolas-grekas: Is there any relation to per-host connections limit?
Because I think that's the most useful limit - per-authority actually.
Global connection limits aren't much needed I think (but I would happily like to know why I'm wrong, if I am!)

trowski: The connection-limit is per authority. I could use a static constructor like `LimitedConnectionPool` to make this clearer.

src/Connection/FiniteConnectionPool.php
```diff
@@ -41,10 +41,23 @@ final class FiniteConnectionPool implements ConnectionPool
...
-    public function __construct(int $maxConnections, ?ConnectionFactory $connectionFactory = null)
+    /**
+     * Create a connection pool that limits the number of connections per authority to $maxConnections.
+     *
+     * @param int                    $maxConnections
+     * @param ConnectionFactory|null $connectionFactory
+     *
+     * @return self
+     */
+    public static function perAuthority(int $maxConnections, ?ConnectionFactory $connectionFactory = null): self
```
nicolas-grekas: byHost for parity with LimitedConnectionPool::byHost? or byAuthority?

trowski: `byAuthority` would probably be best, since it is different than `LimitedConnectionPool::byHost`, as `FiniteConnectionPool::byAuthority(1)` would allow a connection to `http://example.com:80` and `https://example.com:443`.

kelunik:
Why SingleStreamConnectionPool? It isn't a single stream, but the concurrency you specify in the semaphore.

trowski:
@kelunik Yes, was thinking of a mutex not a semaphore. I renamed it to `StreamLimitingConnectionPool` instead. I can revert that commit if we just want to keep the original name.

kelunik:
@nicolas-grekas Could you clarify which limit you actually intend to use? I guess you want something like 6 connections for HTTP/1 and one connection for HTTP/2?

nicolas-grekas:
Yes correct. Using up to 6 connections on h2 would work too, but I don't know which algo should select opening a new connection vs a new stream. Dunno also how curl does it.

kelunik:
But isn't the limit also there to limit concurrent requests, which is then basically no longer the case for HTTP/2 connections?

nicolas-grekas:
Yes that's true. Yet browsers still open several connections to work around TCP throttling, which is the drawback of h2, the reason why h3 exists.

trowski:
Pushed a commit that separates the HTTP/2 and HTTP/1.x connection limits in the pool. Default is 6 HTTP/1.x and 1 HTTP/2. Does this make sense for most use-cases?

src/Connection/FiniteConnectionPool.php
```diff
@@ -0,0 +1,336 @@
...
+    private $openConnectionCount = 0;
+
+    /** @var callable */
+    private $requestToKeyMapper;
+
+    /**
+     * Create a connection pool that limits the number of connections per authority.
+     *
+     * @param int                    $maxHttp1Connections Maximum number of HTTP/2 connections allowed to a single authority.
+     * @param int                    $maxHttp2Connections Maximum number of HTTP/1.x connections allowed to a single authority.
```
nicolas-grekas: descriptions are swapped in the file :)

src/Connection/FiniteConnectionPool.php
```diff
@@ -0,0 +1,336 @@
...
+        callable $requestToKeyMapper,
+        int $maxHttp1Connections,
+        int $maxHttp2Connections,
+        ?ConnectionFactory $connectionFactory = null
+    ) {
+        if ($maxHttp1Connections < 1) {
+            throw new \Error('The number of max HTTP/1.x connections per key must be greater than 0');
+        }
+
+        if ($maxHttp1Connections < 1) {
```
nicolas-grekas: Http2

src/Connection/FiniteConnectionPool.php
trowski: Oops, and then I copied and pasted it everywhere! 😝 Fixed.

src/Connection/FiniteConnectionPool.php
trowski: Fixed.

kelunik:
> Is there a way to race for "new connection is ready" vs "existing connection becomes available" and have this logic decide which stream is used for the next request?

Yes, that should be possible.

nicolas-grekas:
I hope my last comment didn't block progress on this one :)

trowski:
@nicolas-grekas Maybe a little. 😏

I have another branch that extends this PR, what do you think? Can you give it a try? If that approach works well, I'll merge it into this PR.

nicolas-grekas:
I just checked out this PR and I confirm it fixes the perf issue, which means connections are now much better reused.
Since Symfony has a single setting for the number of maximum connections to open I'm creating the pool using:
`$pool = CountLimitingConnectionPool::byAuthority($maxHostConnections, $maxHostConnections, $factory);`

I tried using `1` for the number of HTTP/2 connections but that changes nothing as far as perf is concerned on my use case (actually it's a bit slower).

I think there should be only one single `maxHttpConnections` argument here too.

nicolas-grekas:
PR updated on Symfony's side, now using a single $maxHostConnections setting, all green. Merge + tag pending, then we'll be able to merge on Symfony's side \o/

src/Connection/CountLimitingConnectionPool.php
```diff
@@ -0,0 +1,306 @@
...
+        $stream = yield $this->getStreamFrom($connection, $request);
+
+        if ($stream === null) {
+            // Reused connection did not have an available stream for the given request.
+            $connection = yield $connectionPromise; // Wait for new connection request instead.
+
+            $stream = yield $this->getStreamFrom($connection, $request);
+
+            // New connection should always return a stream for the request.
+            \assert($stream instanceof Stream);
```
kelunik: This might fail, because other requests can subscribe to the same connection promise and we're no longer the first one to subscribe here, so another waiting requests might be "quicker" (executed first) and take the remaining streams.

src/Connection/CountLimitingConnectionPool.php
```diff
@@ -0,0 +1,306 @@
...
+        }
+
+        return [$connection, $stream];
+    }
+
+    private function getStreamFrom(Connection $connection, Request $request): Promise
```
kelunik: I think this should rather be named `getStreamFromConnection` or `getConnectionStream` to make it more visually distinct from `getStreamFor`.

src/Connection/CountLimitingConnectionPool.php
```diff
@@ -0,0 +1,306 @@
...
+    {
+        if (!\array_intersect($request->getProtocolVersions(), $connection->getProtocolVersions())) {
+            return new Success; // Connection does not support any of the requested protocol versions.
+        }
+
+        return $connection->getStream($request);
+    }
+
+    private function shouldMakeNewConnection(string $uri): bool
```
kelunik: I think `isAdditionalConnectionAllowed` would be a better name.

src/Connection/CountLimitingConnectionPool.php
```diff
@@ -0,0 +1,306 @@
...
+        }
+    }
+
+    private function dropConnection(string $uri, int $connectionId): void
+    {
+        unset($this->connections[$uri][$connectionId]);
+
+        if (empty($this->connections[$uri])) {
+            unset($this->connections[$uri], $this->waitForPriorConnection[$uri]);
+        }
```
kelunik: If we drop a connection, shouldn't we do the same thing as `release` then?

src/Connection/CountLimitingConnectionPool.php
```diff
@@ -0,0 +1,306 @@
...
+use Amp\Http\Client\Internal\ForbidSerialization;
+use Amp\Http\Client\Request;
+use Amp\Http\Client\Response;
+use Amp\MultiReasonException;
+use Amp\Promise;
+use Amp\Success;
+use function Amp\call;
+use function Amp\coroutine;
+
+final class CountLimitingConnectionPool implements ConnectionPool
```
kelunik: Should this be `ConnectionLimitingConnectionPool` (vs. `StreamLimitingConnectionPool`)? I guess that's a stupid name.

src/Connection/CountLimitingConnectionPool.php
trowski: Yep, that could happen I suppose. Fixed.

src/Connection/CountLimitingConnectionPool.php
trowski: Not exactly, as that's suppose to be done when a connection is available for another request. I renamed the method accordingly.

src/Connection/CountLimitingConnectionPool.php
trowski: I thought of that name too, but yes, it's a stupid name. `ConnectionCountLimitingPool`?

trowski: And maybe `StreamCountLimitingPool`?

kelunik: Maybe just `StreamLimitingPool` and `ConnectionLimitingPool`?

trowski: Works for me. Done.

kelunik:
@nicolas-grekas We're done. \o/

PR #254, opened 2020-01-17
Await events @ HTTP/2

kelunik:
Not sure whether HTTP/1 or HTTP/2 is better in calling `startReceivingResponse`, technically, these are indeed two responses that are started, but one is only intermediate. Should it trigger for the intermediate, the actual response or both?


trowski:
The actual response IMO.

PR #258, opened 2020-03-03
Close HTTP/1.x connection if stream unavailable

### LINKED ISSUE #256: Timeout when using ConnectionLimitingPool
Sorry for the heavy reproducer - this happens on the Symfony test suite, see https://github.com/symfony/symfony/pull/35924#issuecomment-593897611 for details and reproducer.

trowski:
Fixes #256.

### DIFF
--- a/src/Connection/Http1Connection.php
+++ b/src/Connection/Http1Connection.php
@@ -73,9 +73,6 @@ final class Http1Connection implements Connection
     /** @var int */
     private $estimatedClose;
 
-    /** @var bool */
-    private $explicitTimeout = false;
-
     /** @var SocketAddress */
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
Cool, that was fast thank you. I confirm it fixes the reproducer on my laptop also :)

kelunik:
> What I don't like about your solution is that it will kill the first connection for a POST, even if the limit is high enough to open a new connection and reuse the existing one for GETs

^ Reposting my comment from our chat, so we don't forget about it. This PR will make performance sub-optimal for those situations.

trowski:
Closing in favor of #259.

PR #259, opened 2020-03-04
Tally active requests in pool to close unsuited, idle connections

### LINKED ISSUE #256: Timeout when using ConnectionLimitingPool

trowski:
Fixes #256.

### DIFF
--- a/src/Connection/ConnectionLimitingPool.php
+++ b/src/Connection/ConnectionLimitingPool.php
@@ -176,8 +176,9 @@ final class ConnectionLimitingPool implements ConnectionPool
                 $stream = yield $this->getStreamFromConnection($connection, $request);
 
                 if ($stream === null) {
+                    $connectionId = \spl_object_id($connection);
                     if (!$this->isAdditionalConnectionAllowed($uri)
-                        && $this->activeRequestCounts[\spl_object_id($connection)] === 0
+                        && ($this->activeRequestCounts[$connectionId] ?? 0) === 0
                     ) {
                         // No additional connections allowed, but this connection is idle and unsuited for this request.
                         $connection->close();
--- a/src/Connection/ConnectionLimitingPool.php
+++ b/src/Connection/ConnectionLimitingPool.php
@@ -176,17 +176,7 @@ final class ConnectionLimitingPool implements ConnectionPool
                 $stream = yield $this->getStreamFromConnection($connection, $request);
 
                 if ($stream === null) {
-                    $connectionId = \spl_object_id($connection);
-
-                    \assert(
-                        !isset($this->activeRequestCounts[$connectionId])
-                        || $this->activeRequestCounts[$connectionId] >= 0
-                    );
-
-                    if (!$this->isAdditionalConnectionAllowed($uri)
-                        && ($this->activeRequestCounts[$connectionId] ?? 0) === 0
-                    ) {
-                        // No additional connections allowed, but this connection is idle and unsuited for this request.
+                    if (!$this->isAdditionalConnectionAllowed($uri) && $this->isConnectionIdle($connection)) {
                         $connection->close();
                         break;
                     }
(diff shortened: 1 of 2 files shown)


src/Connection/ConnectionLimitingPool.php
```diff
@@ -170,6 +176,14 @@ private function getStreamFor(string $uri, Request $request, CancellationToken $
                 $stream = yield $this->getStreamFromConnection($connection, $request);
 
                 if ($stream === null) {
+                    if (!$this->isAdditionalConnectionAllowed($uri)
+                        && $this->activeRequestCounts[\spl_object_id($connection)] === 0
```
kelunik: Maybe add an assertion that `$this->activeRequestCounts[\spl_object_id($connection)] >= 0`? I think it also needs a `?? 0`, because the connection might close during `getStreamFromConnection` above.

src/Connection/Http1Connection.php
```diff
@@ -540,9 +537,6 @@ private function determineKeepAliveTimeout(Response $response): int
         $params = Http\createFieldValueComponentMap(Http\parseFieldValueComponents($response, 'keep-alive'));
 
         $timeout = (int) ($params['timeout'] ?? $this->priorTimeout);
-        if (isset($params['timeout'])) {
-            $this->explicitTimeout = true;
-        }
```
kelunik: I think we need to keep the `explicitTimeout` flag. With the current change, all connections without timeout will close after the grace period, even if it's entirely safe to reuse them for GET requests.

src/Connection/Http1Connection.php
```diff
@@ -177,7 +177,7 @@ private function free(): Promise
 
     private function hasStreamFor(Request $request): bool
     {
-        $connectionUnlikelyToClose = $this->explicitTimeout && $this->getRemainingTime() > $this->timeoutGracePeriod;
+        $connectionUnlikelyToClose = $this->getRemainingTime() > $this->timeoutGracePeriod;
```
kelunik: `getRemainingTime()` will return `$this->timeoutGracePeriod`, which won't be greater than `$this->timeoutGracePeriod`, no?

trowski: Oops, yes. What should the timeout be then if no explicit timeout is given?

kelunik: The hard connection close timeout will be `MAX_KEEP_ALIVE_TIMEOUT`. If the request is idempotent, the timeout is `MAX_KEEP_ALIVE_TIMEOUT` as well. If it is non-idempotent, it's `$this->timeoutGracePeriod`.

If there's an explicit timeout, it's `$remainingTime > $this->timeoutGracePeriod` for non-idempotent requests, idempotent requests don't need to check the remaining time.

trowski: Yeah, ok, fixed. I should have been just comparing to 0 since I changed how `getRemainingTime()` worked.

trowski:
@nicolas-grekas This PR is a more robust solution, as @kelunik pointed out that the same situation could happen with multiple HTTP/2 requests, followed by an HTTP/1.x request (and certainly in other scenarios). This PR now closes connections in the pool if they are idle but unable to fulfill a new request.

This PR also relaxes reusing an HTTP/1.x keep-alive connection *without* an explicit timeout (which was the issue that caused the timeouts you saw) for a `POST` or `PUT` request as long as it is within the given connection grace period (2 seconds by default).

src/Connection/ConnectionLimitingPool.php
```diff
@@ -170,6 +176,21 @@ private function getStreamFor(string $uri, Request $request, CancellationToken $
...
+
+                    \assert(
+                        !isset($this->activeRequestCounts[$connectionId])
+                        || $this->activeRequestCounts[$connectionId] >= 0
+                    );
+
+                    if (!$this->isAdditionalConnectionAllowed($uri)
+                        && ($this->activeRequestCounts[$connectionId] ?? 0) === 0
+                    ) {
+                        // No additional connections allowed, but this connection is idle and unsuited for this request.
```
kelunik: Instead of the comment, we could refactor the second condition of the `if` into an `$this->isIdle($connection)` method.

PR #262, opened 2020-03-10
Fix transfer timeout for HTTP/2

### LINKED ISSUE #260: Body transfer timeout possibly not enforced correctly
More details will follow.

kelunik:
Fixes #260.

### DIFF
--- a/src/Connection/Internal/Http2ConnectionProcessor.php
+++ b/src/Connection/Internal/Http2ConnectionProcessor.php
@@ -1307,10 +1307,10 @@ final class Http2ConnectionProcessor implements Http2Processor
 
         if ($stream->responsePending || $stream->body || $stream->trailers) {
             $exception = $exception ?? new ClientHttp2StreamException(
-                    \sprintf("Stream %d closed unexpectedly", $streamId),
-                    $streamId,
-                    Http2Parser::INTERNAL_ERROR
-                );
+                \sprintf("Stream %d closed unexpectedly", $streamId),
+                $streamId,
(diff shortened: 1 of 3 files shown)


src/Connection/Internal/Http2ConnectionProcessor.php
```diff
@@ -905,29 +918,34 @@ public function request(Request $request, CancellationToken $cancellationToken,
...
             );
 
-            if ($request->getTransferTimeout() > 0) {
-                // Cancellation token combined with timeout token should not be stored in $stream->cancellationToken,
-                // otherwise the timeout applies to the body transfer and pushes.
-                $cancellationToken = new CombinedCancellationToken(
-                    $cancellationToken,
-                    new TimeoutCancellationToken($request->getTransferTimeout())
-                );
-            }
```
kelunik: Seems like this commend has been introduced by you, @trowski: 93c2e6e5baab68558deb8c0c3e7b8131726ba1db

trowski: I think my assumption at that time was that the transfer timeout only applied to the headers and that there was a separate timeout for transferring the response body.

kelunik:
@trowski I'm not sure how pushes should be handled if the parent is cancelled / times out. Should there be a difference in handling? I guess for cancelling, cancelling the pushes makes sense, because they're essentially children of the parent stream.

trowski:
I agree that a push is essentially a child of the parent request. If the parent is cancelled for any reason (timeout, etc.), all the pushes associated should be cancelled.

PR #263, opened 2020-03-11
Add inactivity timeout

trowski:
This provides a separate timeout while waiting for the response or streaming the body. If no data is received for the response within the given number of milliseconds, the request fails similarly to the transfer timeout.

### DIFF
--- a/src/Connection/Http1Connection.php
+++ b/src/Connection/Http1Connection.php
@@ -112,10 +112,7 @@ final class Http1Connection implements Connection
 
     public function close(): Promise
     {
-        if ($this->socket) {
-            $this->socket->close();
-        }
-
+        $this->socket->close();
         return $this->free();
     }
 
@@ -156,7 +153,6 @@ final class Http1Connection implements Connection
 
     private function free(): Promise
     {
-        $this->socket = null;
         $this->lastUsedAt = 0;
 
         if ($this->timeoutWatcher !== null) {
@@ -280,14 +276,13 @@ final class Http1Connection implements Connection
 
         $start = getCurrentTime();
         $timeout = $request->getInactivityTimeout();
-        if ($timeout <= 0) {
-            $timeout = self::MAX_KEEP_ALIVE_TIMEOUT * 1000;
-        }
 
         try {
-            while (null !== $chunk = yield Promise\timeout($this->socket->read(), $timeout)) {
-                parseChunk:
-                $response = $parser->parse($chunk);
+            while (null !== $chunk = yield $timeout > 0
+                    ? Promise\timeout($this->socket->read(), $timeout)
+                    : $this->socket->read()
+            ) {
+                parseChunk: $response = $parser->parse($chunk);
                 if ($response === null) {
                     continue;
                 }
@@ -395,10 +390,13 @@ final class Http1Connection implements Connection
                                     if (!$backpressure instanceof Success) {
                                         yield $this->withCancellation($backpressure, $bodyCancellationToken);
                                     }
-                                } while (null !== $chunk = yield Promise\timeout($this->socket->read(), $timeout));
+                                } while (null !== $chunk = yield $timeout > 0
+                                    ? Promise\timeout($this->socket->read(), $timeout)
+                                    : $this->socket->read()
+                                );
                             } catch (PromiseTimeoutException $e) {
                                 $this->close();
-                                throw new TimeoutException('Receiving the response timed out due to inactivity', 0, $e);
+                                throw new TimeoutException('Inactivity timeout exceeded, took longer than ' . $timeout . ' ms', 0, $e);
                             }
 
                             $originalCancellation->throwIfRequested();
@@ -461,6 +459,9 @@ final class Http1Connection implements Connection
                 \strlen($parser->getBuffer()),
                 getCurrentTime() - $start
             ));
+        } catch (HttpException $e) {
+            $this->close();
+            throw $e;
         } catch (PromiseTimeoutException $e) {
             $this->close();
             throw new TimeoutException('Receiving the response timed out due to inactivity', 0, $e);
--- a/src/Connection/Http1Connection.php
+++ b/src/Connection/Http1Connection.php
@@ -32,9 +32,9 @@ use Amp\Socket\SocketAddress;
 use Amp\Socket\TlsInfo;
 use Amp\Success;
 use Amp\TimeoutCancellationToken;
+use Amp\TimeoutException as PromiseTimeoutException;
 use function Amp\asyncCall;
 use function Amp\call;
-use function Amp\getCurrentTime;
 
 /**
  * Socket client implementation.
@@ -92,7 +92,7 @@ final class Http1Connection implements Connection
         $this->remoteAddress = $socket->getRemoteAddress();
         $this->tlsInfo = $socket->getTlsInfo();
         $this->timeoutGracePeriod = $timeoutGracePeriod;
-        $this->lastUsedAt = getCurrentTime();
+        $this->lastUsedAt = Loop::now();
     }
 
     public function __destruct()
@@ -279,14 +279,20 @@ final class Http1Connection implements Connection
         $parser = new Http1Parser($request, $bodyCallback, $trailersCallback);
 
         $start = getCurrentTime();
+        $timeout = $request->getInactivityTimeout();
+        if ($timeout <= 0) {
+            $timeout = self::MAX_KEEP_ALIVE_TIMEOUT * 1000;
+        }
 
         try {
-            while (null !== $chunk = yield $this->socket->read()) {
+            while (null !== $chunk = yield Promise\timeout($this->socket->read(), $timeout)) {
                 parseChunk:
                 $response = $parser->parse($chunk);
                 if ($response === null) {
                     continue;
                 }
+
+                $this->lastUsedAt = Loop::now();
 
                 $status = $response->getStatus();
 
@@ -362,6 +368,7 @@ final class Http1Connection implements Connection
                     $readingCancellation,
                     $bodyCancellationToken,
                     $stream,
+                    $timeout,
                     &$backpressure,
                     &$trailers
                 ) {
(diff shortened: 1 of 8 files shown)


src/Connection/Http1Connection.php
```diff
@@ -92,7 +92,7 @@ public function __construct(EncryptableSocket $socket, int $timeoutGracePeriod =
         $this->remoteAddress = $socket->getRemoteAddress();
         $this->tlsInfo = $socket->getTlsInfo();
         $this->timeoutGracePeriod = $timeoutGracePeriod;
-        $this->lastUsedAt = getCurrentTime();
+        $this->lastUsedAt = Loop::now();
```
kelunik: Why `Loop::now()` instead of `getCurrentTime()`?

src/Connection/Http1Connection.php
```diff
@@ -276,16 +276,22 @@ private function readResponse(
 
         $parser = new Http1Parser($request, $bodyCallback, $trailersCallback);
 
-        $start = getCurrentTime();
+        $start = Loop::now();
+        $timeout = $request->getInactivityTimeout();
+        if ($timeout <= 0) {
+            $timeout = self::MAX_KEEP_ALIVE_TIMEOUT * 1000;
```
kelunik: Why this fallback instead of letting the user disable the timeout entirely?

src/Connection/Http1Connection.php
```diff
@@ -372,18 +379,23 @@ private function readResponse(
...
+                                        break;
+                                    }
+
+                                    if (!$backpressure instanceof Success) {
+                                        yield $this->withCancellation($backpressure, $bodyCancellationToken);
+                                    }
+                                } while (null !== $chunk = yield Promise\timeout($this->socket->read(), $timeout));
+                            } catch (PromiseTimeoutException $e) {
+                                $this->close();
+                                throw new TimeoutException('Receiving the response timed out due to inactivity', 0, $e);
```
kelunik: This should probably include the configured timeout value like all other timeout error messages.

src/Connection/Internal/Http2Stream.php
```diff
@@ -87,12 +88,19 @@ final class Http2Stream
...
+    /** @var int */
+    public $lastActivityAt;
+
     public function __construct(
         int $id,
         Request $request,
         Stream $stream,
         CancellationToken $cancellationToken,
         CancellationToken $originalCancellation,
+        ?string $watcher,
```
kelunik: Why is it nullable?

src/Request.php
```diff
@@ -49,7 +49,10 @@ private static function clone($value)
     private $tlsHandshakeTimeout = 10000;
 
     /** @var int */
-    private $transferTimeout = 10000;
+    private $transferTimeout = 60000;
```
kelunik: I don't see any reason to change the default here.

src/Request.php
```diff
@@ -49,7 +49,10 @@ private static function clone($value)
...
+
+    /** @var int */
+    private $inactivityTimeout = 5000;
```
kelunik: I think I'd set this equal to the `transferTimeout` for now, so `10000`, keeping everything as is for existing users.

src/Connection/Internal/Http2ConnectionProcessor.php
```diff
@@ -1564,4 +1574,41 @@ private function increaseStreamWindow(Http2Stream $stream): void
...
+                    \pack("N", Http2Parser::CANCEL)
+                );
+
+                $this->releaseStream($streamId, new TimeoutException('The request timed out due to inactivity'));
+                return;
+            }
+
+            $delay = $timeout - ($now - $stream->lastActivityAt);
+
+            $stream->watcher = $this->createStreamInactivityWatcher($streamId, $timeout, $delay);
```
kelunik: Shouldn't we disable / re-enable the watcher on activity instead?

src/Connection/Http1Connection.php
trowski: Only because `Loop::now()` caches it during a tick.

src/Connection/Http1Connection.php
trowski: Meant to change – was being lazy so I could always use `Promise\timeout()`.

src/Connection/Internal/Http2Stream.php
trowski: If there is no inactivity timeout, no watcher is created.

src/Request.php
trowski: I was thinking for better UX when requesting a large file. Having all the timeouts the same isn't really helpful.

src/Request.php
trowski: For BC I suppose, though as I said above, different timeouts seems more practical IMO.

src/Connection/Internal/Http2ConnectionProcessor.php
trowski: Yeah, that's easier. Was thinking it would happen a lot, but no different than how the read watcher behaves.

src/Connection/Http1Connection.php
kelunik: I think this is rather a disadvantage, because timeouts will be shorter on a busy loop then.

src/Request.php
kelunik: I think downloading large files is rather uncommon compared to making API calls where large timeouts aren't a good idea.

src/Connection/Http1Connection.php
trowski: I switched it to `getCurrentTime()`.

src/Request.php
kelunik: Yes, we can definitely change those in v5.

src/Request.php
trowski: Already reverted the timeouts back to pre-PR values.

src/Request.php
trowski: Sounds good.

enumag: What timeouts would you recommend? A docblock comment like "The default value will be changed to xxx in v5" would be nice.

src/Connection/Http1Connection.php
```diff
@@ -445,8 +456,15 @@ private function readResponse(
...
                 getCurrentTime() - $start
             ));
-        } catch (StreamException $e) {
-            throw new SocketException('Receiving the response headers failed: ' . $e->getMessage());
+        } catch (HttpException $e) {
+            $this->close();
+            throw $e;
+        } catch (PromiseTimeoutException $e) {
+            $this->close();
+            throw new TimeoutException('Receiving the response timed out due to inactivity', 0, $e);
```
kelunik, line 464: The exception message should also have the timeout in the message here.

test/Connection/Http1ConnectionTest.php
```diff
@@ -124,6 +125,66 @@ public function testUpgrade(): \Generator
...
+
+        $connection = new Http1Connection($client);
+
+        $request = new Request('http://localhost');
+        $request->setInactivityTimeout(500);
+
+        /** @var Stream $stream */
+        $stream = yield $connection->getStream($request);
+
+        $server->write("HTTP/1.1 200 Continue\r\nConnection: keep-alive\r\nContent-Length: 8\r\n\r\ntest");
```
kelunik, line 173: Should this test send two chunks to differentiate it from the transfer timeout test case?

src/Connection/Http1Connection.php
trowski: Both locations fixed.

test/Connection/Http1ConnectionTest.php
trowski: Good idea, updated.

src/Request.php
kelunik: @enumag It mostly depends on the situation you're using the client in. It might be a good idea to give some pre-configurations as part of `HttpClientBuilder`.

kelunik:
@trowski Could you resolve the conflicts?

src/Connection/Http1Connection.php
```diff
@@ -156,7 +154,6 @@ public function getStream(Request $request): Promise
 
     private function free(): Promise
     {
-        $this->socket = null;
```
kelunik, line 159: We have several null checks now, so this should be unnecessary. Or should we revert the null checks?

trowski: I think this was a rebase mistake. I didn't mean to remove that.

src/Connection/Http1Connection.php
```diff
@@ -374,24 +382,28 @@ private function readResponse(
...
-                                    break;
-                                }
-
-                                if (!$backpressure instanceof Success) {
-                                    yield $this->withCancellation($backpressure, $bodyCancellationToken);
-                                }
-
-                                /** @psalm-suppress TypeDoesNotContainNull */
-                                if ($this->socket === null) {
-                                    throw new SocketException('Socket closed prior to response completion');
```
kelunik, line 392: Seems like this also got accidentally removed due to rebasing?

src/Connection/Internal/Http2ConnectionProcessor.php
```diff
@@ -711,6 +714,7 @@ public function handlePriority(int $streamId, int $parentId, int $weight): void
         }
 
         $stream = $this->streams[$streamId];
+        $stream->resetInactivityWatcher();
```
kelunik, line 717: Should priority changes really reset the inactivity watcher?

src/Connection/Internal/Http2ConnectionProcessor.php
```diff
@@ -217,6 +217,7 @@ public function handleStreamWindowIncrement(int $streamId, int $windowSize): voi
...
```
kelunik, line 220: Same here, not sure whether this should reset the inactivity watcher.

src/Connection/Http1Connection.php
trowski: Yep, missed adding that back. Fixed.

src/Connection/Internal/Http2ConnectionProcessor.php
trowski: Nah, probably not. Removed.

src/Connection/Http1Connection.php
```diff
@@ -374,24 +382,33 @@ private function readResponse(
...
+                                /** @psalm-suppress PossiblyNullReference */
+                                do {
+                                    /** @noinspection CallableParameterUseCaseInTypeContextInspection */
+                                    $parser->parse($chunk);
+                                    /** @noinspection NotOptimalIfConditionsInspection */
+                                    if ($parser->isComplete()) {
+                                        break;
+                                    }
+
+                                    /** @psalm-suppress TypeDoesNotContainNull */
```
kelunik, line 395: I guess the suppression here is also a rebase artifact from the `if` below?

trowski: Yeah, funny that I only got that line.

PR #267, opened 2020-04-30
Make conflict with amphp/file lax

enumag:
I'm trying to use `"amphp/file": "dev-master as 1.0.1",` in my composer.json because of https://github.com/amphp/file/commit/7dc65ffc03d236bf0465b83ed7767f043f502873 but composer refuses to install it:

```
  Problem 1
    - amphp/file dev-master conflicts with amphp/http-client[v4.2.2].
    - amphp/http-client v4.2.2 conflicts with amphp/file[dev-master].
    - amphp/file dev-master conflicts with amphp/http-client[v4.2.2].
    - Installation request for amphp/file dev-master as 1.0.1 -> satisfiable by amphp/file[dev-master].
    - Installation request for amphp/http-client (locked at v4.2.2, required as ^4.0) -> satisfiable by amphp/http-client[v4.2.2].
```

Now there are two ways to fix this:

1. Add [branch alias](https://getcomposer.org/doc/articles/aliases.md#branch-alias) to amphp/file to let composer know that dev-master is 1.x - then I could depend on `^1.0@dev`. (I'd recommend this for all repos.)
2. Remove the conflict with amphp/file v2 here. It doesn't make sense at all since v2 doesn't even exist and the constraint blocks dev versions.

Personally I'd recommend doing both.

Please fix this ASAP.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -64,7 +64,7 @@
         }
     },
     "conflict": {
-        "amphp/file": "<0.2 || >=2"
+        "amphp/file": "<0.2"
     },
     "scripts": {
         "check": [


kelunik:
> 2\. Remove the conflict with amphp/file v2 here. It doesn't make sense at all since v2 doesn't even exist and the constraint blocks dev versions.

It makes as much sense as using `"require": { "amphp/file": "^1" }` instead of ``"require": { "amphp/file": ">=1" }`. It's basically an optional `"require"`.

kelunik:
I've added the branch alias to `amphp/file`.

enumag:
@kelunik I tried again but unfortunately it seems it didn't help. Composer is still refusing to install it.

kelunik:
Works fine for me, how does your `composer.json` look like?

enumag:
Yeah, minimal use-case works for me too. But it breaks as soon as you add anything (even totally unrelated) into require-dev:

```
{
    "require": {
        "amphp/file": "^1.0@dev",
        "amphp/http-client": "^4.2"
    },
    "require-dev": {
        "psr/log": "*"
    }
}
```

```                                                                                                      
  [Composer\DependencyResolver\SolverProblemsException]                                               
  Problem 1                                                                                           
      - Installation request for amphp/file ^1.0@dev -> satisfiable by amphp/file[1.x-dev].           
      - amphp/file dev-master conflicts with amphp/http-client[v4.2.2].                               
      - remove amphp/file 1.x-dev|keep amphp/file dev-master                                          
      - Installation request for amphp/http-client ^4.2 -> satisfiable by amphp/http-client[v4.2.2].                                                                                                     
```

Honestly I think it's a composer bug. Can you please tell me if this happens for you too?

kelunik:
Confirmed, this pretty much looks like a composer bug.

enumag:
Thanks... could you make a stable release of amphp/file with https://github.com/amphp/file/commit/7dc65ffc03d236bf0465b83ed7767f043f502873? I'd like to avoid using workarounds.

enumag:
@kelunik As you might have noticed the bug won't be fixed in Composer 1.x. Would you consider removing the >=2 conflict for now until Composer 2.0 is stable and commonly used?

maxwagner-dev:
Got the same error as @enumag . My workaround currently is setting amphp/file strict to version v1.0.0 instead of ^v1.0.0. 
Maybe you can consider implement this Pull-Request to avoid any further problems, until version 2 of amphp/file is out.

kelunik:
Using `amphp/http-client => ^4@dev` should work for now then. :-)

PR #269, opened 2020-05-08
Error if request URI provides relative path

remorhaz:


remorhaz:
The build is failing but it seems to have been failing for some time. Tests are segfaulting for some reason under 7.2 even locally (I failed to understand why) but run okay on 7.3 and 7.4. I've re-checked that at least my tests are running okay on all three versions.

test/Connection/Http1ConnectionTest.php
```diff
@@ -194,6 +198,74 @@ public function testInactivityTimeout(): \Generator
...
+        $uri->method('getHost')->willReturn('localhost');
+        $uri->method('getQuery')->willReturn('');
+        $uri->method('getFragment')->willReturn('');
+        $uri->method('withScheme')->willReturnSelf();
+        $uri->method('withUserInfo')->willReturnSelf();
+        $uri->method('withHost')->willReturnSelf();
+        $uri->method('withPort')->willReturnSelf();
+        $uri->method('withQuery')->willReturnSelf();
+        $uri->method('withFragment')->willReturnSelf();
+        $uri->method('__toString')->willReturn('http://localhost/foo');
```
kelunik, line 219: Do we need a mock here or can we just use `league/uri`?

remorhaz: We need mock because `league/uri` doesn't allow relative paths and I don't think it's a good idea to require-dev another implementation that does allow them (`laminas/laminas-diactoros`, for example).

kelunik: That's good to know, thanks! I wouldn't have a problem with requiring `laminas/laminas-diactoros` in `require-dev`.

kelunik:
If you rebase onto `master` or merge `master`, the tests should be working fine again. I've skipped the tests causing segfaults on PHP 7.2 and 7.3.

remorhaz:
> The same check needs to be done in `Http2Connection`, should we put the check into `DefaultConnectionFactory` or duplicate it into `Http2Connection`?

Hmm, I need to take a closer look...

test/Connection/Http1ConnectionTest.php
remorhaz: That's not a big problem, but they can decide to change this behavior in future - I think it's better not to rely on it. Anyway, it's just one small negative corner-case.

remorhaz:
> should we put the check into `DefaultConnectionFactory` or duplicate it into `Http2Connection`?

I think it's wise to make checks just before contract protects us; and in our case contract of `UriInterface` doesn't protect us from this particular problem, so I vote for duplication. Relying on `DefaultConnectionFactory` is out of contact at all, I believe that code shouldn't rely on some checks made by external code.

remorhaz:
Seems that some tests are flaky.

kelunik:
Thanks!

Issue #266, opened 2020-04-28
PSR-7


enumag:
Do you have / know about any existing implementation that would allow me to transform a [PSR-7](https://www.php-fig.org/psr/psr-7/) request into `Amp\Http\Client\Request` and later a `Amp\Http\Client\Response` into a PSR-7 response?

kelunik:
I don't know about anything like that, but we could probably easily add it.

nicolas-grekas:
https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpClient/Psr18Client.php

This will do the job and allow mapping this to that.
Note that PSR-7 is mostly not compatible with any async feature so there is little better you can do than use PSR-18.

remorhaz:
That would be a nice feature, sometimes there's a need to pass messages to PSR-compatible code or receive it from it, and I have to implement adapters by myself. It's an easy task, but it's better to have it done by library.

kelunik:
@remorhaz If you already have the code, a PR is welcome. ;-)

remorhaz:
I guess it's better to implement it as a part of `amphp/http` (consider moving the issue). I will prepare the PR soon.

remorhaz:
Oops, seems that `Message` class is too abstract, it doens't even contain a body. In this case let's move from particulars to generals.

kelunik:
Yes, it needs to be implemented here instead.

kelunik:
See https://github.com/amphp/http-client-psr7.

PR #270, opened 2020-05-11
Add PSR-7 request adapter (#266)

remorhaz:
This PR shows partial job, just to sync expectations. It only implements converting requests between Amp and PSR-7 and vice versa.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -69,7 +69,8 @@
         }
     },
     "conflict": {
-        "amphp/file": "<0.2"
+        "amphp/file": "<0.2",
+        "psr/http-factory": "<1 || >=2"
     },
     "scripts": {
         "check": [
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,8 @@
         "amphp/socket": "^1",
         "amphp/sync": "^1.3",
         "league/uri": "^6",
-        "psr/http-message": "^1"
+        "psr/http-message": "^1",
+        "psr/http-factory": "^1"
     },
     "require-dev": {
         "ext-json": "*",
--- a/src/Body/PsrStreamBody.php
+++ b/src/Body/PsrStreamBody.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 namespace Amp\Http\Client\Body;
 
 use Amp\ByteStream\InputStream;
@@ -13,7 +11,6 @@ use Psr\Http\Message\StreamInterface;
 
 final class PsrStreamBody implements RequestBody
 {
-
     /**
      * @var StreamInterface
      */
--- a/src/Body/PsrStreamBody.php
+++ b/src/Body/PsrStreamBody.php
@@ -1,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Amp\Http\Client\Body;
+
+use Amp\ByteStream\InputStream;
+use Amp\Http\Client\Internal\PsrInputStream;
+use Amp\Http\Client\RequestBody;
+use Amp\Promise;
+use Amp\Success;
+use Psr\Http\Message\StreamInterface;
+
+final class PsrStreamBody implements RequestBody
+{
+
+    /**
+     * @var StreamInterface
+     */
+    private $stream;
+
+    public function __construct(StreamInterface $stream)
+    {
+        $this->stream = $stream;
+    }
+
+    public function getBodyLength(): Promise
+    {
+        return new Success($this->stream->getSize() ?? -1);
+    }
+
+    public function getHeaders(): Promise
+    {
+        return new Success([]);
+    }
+
+    public function createBodyStream(): InputStream
+    {
+        return new PsrInputStream($this->stream);
+    }
+}
--- a/src/Internal/PsrRequestStream.php
+++ b/src/Internal/PsrRequestStream.php
@@ -1,0 +1,192 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Amp\Http\Client\Internal;
+
+use Amp\ByteStream\InputStream;
+use Psr\Http\Message\StreamInterface;
+
+use function Amp\Promise\timeout;
+use function Amp\Promise\wait;
+
+/**
+ * @internal
+ */
+final class PsrRequestStream implements StreamInterface
+{
+    private const DEFAULT_TIMEOUT = 5000;
+
+    /**
+     * @var InputStream
+     */
+    private $stream;
+
+    /**
+     * @var int
+     */
+    private $timeout;
+
+    /**
+     * @var string
+     */
+    private $buffer = '';
+
+    /**
+     * @var bool
+     */
+    private $isEof = false;
+
+    public function __construct(InputStream $stream, int $timeout = self::DEFAULT_TIMEOUT)
+    {
+        $this->stream = $stream;
+        $this->timeout = $timeout;
+    }
+
+    public function __toString()
+    {
+        try {
+            return $this->getContents();
+        } catch (\Throwable $e) {
+            return '';
+        }
+    }
+
+    public function close(): void
+    {
+        unset($this->stream);
+    }
+
+    public function eof(): bool
+    {
+        return $this->isEof;
+    }
+
+    public function tell()
+    {
+        throw new \RuntimeException("Source stream is not seekable");
+    }
+
+    /**
+     * @return null
+     */
+    public function getSize()
+    {
+        return null;
+    }
+
+    /**
+     * @return bool
+     */
+    public function isSeekable()
+    {
+        return false;
+    }
+
+    /**
+     * @param int $offset
+     * @param int $whence
+     * @return void
+     */
+    public function seek($offset, $whence = SEEK_SET)
+    {
+        throw new \RuntimeException("Source stream is not seekable");
+    }
+
+    /**
+     * @return void
+     */
+    public function rewind()
+    {
+        throw new \RuntimeException("Source stream is not seekable");
+    }
+
+    /**
+     * @return bool
+     */
+    public function isWritable()
+    {
+        return false;
+    }
+
+    public function write($string)
+    {
+        throw new \RuntimeException("Source stream is not writable");
+    }
+
+    /**
+     * @param string|null $key
+     * @return array|null
+     */
+    public function getMetadata($key = null)
+    {
+        return isset($key) ? null : [];
+    }
+
+    /**
+     * @return null
+     */
+    public function detach()
+    {
+        unset($this->stream);
+
+        return null;
+    }
+
+    /**
+     * @return bool
+     */
+    public function isReadable()
+    {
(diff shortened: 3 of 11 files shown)


composer.json
```diff
@@ -34,7 +34,8 @@
         "amphp/socket": "^1",
         "amphp/sync": "^1.3",
         "league/uri": "^6",
-        "psr/http-message": "^1"
+        "psr/http-message": "^1",
+        "laminas/laminas-diactoros": "^2.3"
```
nicolas-grekas, line 38: Please move this back to require-dev, it should remain an optional feature.

remorhaz: Now that's impossible because adapter uses Request object from Laminas library. I've tried to find League implementation but they use Laminas in their examples; so Laminas is not optional anymore.

remorhaz: We can avoid requiring concrete PSR-7 implementation, though, if we accept RequestInterface in corresponding method (and then call with* methods). But the problem is body - it's mutable in PSR-7.

nicolas-grekas: None of the added classes are mandatory to use the http-client. If one doesn't use any of them (I don't), one shouldn't be force to install a dead-code dependency. Moving it back to require-dev fixes the issue.

About diactoros, could it be possible to be compatible with nyholm/psr7? That's the lib we should recommend IMHO (I don't know about the policy of the amp project here, I'm just talking for Symfony-related apps.)

remorhaz: Wait, I'll try one idea...

remorhaz: I've got rid of requiring Laminas with injecting PSR-7 `RequestFactoryInterface`. Maybe it would be more convenient to inject it in constructor - but `from*` methods will never use it, so in this case we'll have to split adapter into independent from* and to* parts.

remorhaz: On the other hand, most use cases will require using either  "fromPsrRequest/toPsrResponse" or "toPsrRequest/fromPsrResponse" pair of methods. One of them will require `RequestFactoryInterface`, so I'll better inject it in constructor.

src/Body/PsrStreamBody.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
```
kelunik, line 3: We don't use `strict_types` in any @amphp project, so we shouldn't use it here.

src/Internal/PsrRequestStream.php
```diff
@@ -0,0 +1,192 @@
...
+
+        $data = \substr($this->buffer, 0, $length);
+        $this->buffer = \substr($this->buffer, \strlen($data));
+
+        return $data;
+    }
+
+    private function readFromStream(): string
+    {
+        $data = wait(timeout($this->getOpenStream()->read(), $this->timeout));
```
kelunik, line 162: We can't use `wait`, because it has issues with nesting, see https://github.com/amphp/amp/pull/311 and https://github.com/amphp/amp/pull/313. This means we need to buffer the entire body for these.

composer.json
```diff
@@ -34,7 +34,8 @@
         "amphp/socket": "^1",
         "amphp/sync": "^1.3",
         "league/uri": "^6",
-        "psr/http-message": "^1"
+        "psr/http-message": "^1",
+        "psr/http-factory": "^1"
```
nicolas-grekas, line 38: now you can also remove this one as it's not a dependency at all!
(it's only a dependency of the external part that instantiate the factory)

kelunik: It's a dependency as much as `psr/http-message` is a dependency for `UriInterface`.

nicolas-grekas: Look at the code: `UriInterface` is part of the public signature of `Request::getUri()`.

`RequestFactoryInterface` on the contrary is only part of the constructor of `PsrAdapter`. This means you *can* load the class definition of `PsrAdapter` without any issue. (And you can't for `Request`).

The thing that instantiates `PsrAdapter` must have `psr/http-factory` as a dependency to use the class, but the class and the package themself do not dependent on the `psr/http-factory`.

src/Body/PsrStreamBody.php
```diff
@@ -0,0 +1,41 @@
...
+use Amp\ByteStream\InputStream;
+use Amp\Http\Client\Internal\PsrInputStream;
+use Amp\Http\Client\RequestBody;
+use Amp\Promise;
+use Amp\Success;
+use Psr\Http\Message\StreamInterface;
+
+final class PsrStreamBody implements RequestBody
+{
+
```
enumag, line 16:
```suggestion
```

composer.json
remorhaz: I agree that `psr/http-factory` is much lesser evil than Laminas - it's just interfaces. In case it's unacceptable - the only escape is to extract adapters to separate library.

src/Internal/PsrRequestStream.php
remorhaz: Hmm, that's really bad: the body can be very big in some cases. Is there any workaround that will allow as to read the body by parts? Maybe some non-promise generator?

composer.json
nicolas-grekas: But it is *not* dependency of this package... This should be moved to require-dev as that's the place were it belongs.

src/Internal/PsrRequestStream.php
kelunik: The only one would be writing huge responses to files, I guess.

composer.json
remorhaz: Sorry, but it _is_ dependency, because if we use type in our signature (no matter if as argument or result); as a prrof of my words - require-checker fails without it.

nicolas-grekas: Then that's a bug in require-checker... If you reason about what makes something a dependency, this one is not. Because it's used only in the constructor, it's for sure a dependency of the thing that is passed to the constructor, but it's not a dependency of this specific class.

Trimming what is in "require" is critical to escape from dependency hell, here is an opportunity that we should not miss.

remorhaz: No, it's not a bug. Look: this `RequestFactoryIntreface` is not just an obscure type; we call it's methods and thus we're depending on it's _version_.

I absolutely agree that not every user of `amphp/http-client` needs these adapters at all; we can extract them into the separate library, something like `amphp/http-client-psr7-bridge`. In the same time, supporting a separate library is a work; and PSR packages are small and change rarely. So I delegate the decision to @kelunik - his code, his rules.

src/Internal/PsrRequestStream.php
remorhaz: Is there any chance of solving these issues in the future? In this case I can buffer the entire body for now, but keep the infrastructure to restore current behavior in future.

Is there any (relatively) simple way to reproduce the issues using my code? Maybe I'll think something out.

composer.json
nicolas-grekas: amphp/http-client should not force a dependency on its dependent packages when it's not *required* to have it. This line pushes a dependency that will be dead code to ppl that don't use `PsrAdapter`. And ppl that will use `PsrAdapter` *must* depend on `psr/http-factory` on their own because the are the one that will instantiate that injected object.

If you really want to enforce version boundaries, this should be moved to "conflict" instead.

remorhaz: > This line pushes a dependency that will be dead code to ppl that don't use `PsrAdapter`.

I absolutely agree. But there is no way to detect what parts of a package are _really_ used; thus the only way we have is just assuming that _any_ part of library can be used. Package have to stay consistent.

Is I've said above, the only way to remove this dependency is to move the adapter to another package. "Conflict" is not an option.

nicolas-grekas: Let's say we remove this line. What did we lose in the process? Did we break anything? What does this line provide in the first place?

kelunik: @remorhaz: "conflict" is very much an option, and probably the way to go.

As most users won't use the PSR converter, I'm leaning towards a separate package, as breaking changes seem more likely than for the rest of the code in this package.

src/Internal/PsrRequestStream.php
kelunik: Yes, it's very easy to reproduce, just call that code inside Loop::run() and it'll stop the outer Loop::run(), too.

composer.json
remorhaz: > Let's say we remove this line. What did we lose in the process? Did we break anything?

Of course. If someone uses factory v1 in his code and then upgrades to factory v2 that breaks BC (by renaming factory method, for example), he gets his service broken without any chance of being preserved from it by Composer. That's how the dependency works; just ignoring the dependency doesn't remove it.

"Conflicting" with factory v2+ is pretty the same as requiring v1 - it will not allow to install factory v2 with our library, even if user doesn't use PSR adapters. It's just using wrong mechanism instead of right one, I think; it solves no problems. I hope @kelunik will decide to create a separate package - that's the best way to keep original client's dependencies untouched.

nicolas-grekas: Software design, package principles here, should prepare us for events that are possible. Here the v1/v2 example is unrealistic IMHO when talking about `psr/http-factory`. It's not a good enough reason to force it as a dependency to everyone "just in case". Even adding a "conflict" looks like useless precaution to me.

Anyway, we don't have to agree on this as a separate package might provide a path forward for both our pov.

Thanks for the talk.

kelunik: @nicolas-grekas In the case of `psr/http-factory` the BC break is indeed unlikely, as PSRs don't currently evolve and probably never will, because releasing a v2 on the same package would be a huge pain.

"conflict" seems to be the right choice currently for we're compatible with this API, but don't want an installation dependency.

Indeed, a separate package solves the "problem" entirely. :)

remorhaz:
I've made some changes:

- Added response adapters;
- Adapter returns promises instead of "direct" objects;
- Message bodies are buffered entirely before resolving the message promise (escaping `wait()` issues).
- Minor codestyle fixes, tests added, etc.

trowski:
I agree with @kelunik, this really should be a separate package.

See the PSR adaptor I've been working on for `http-server`: https://github.com/trowski/psr-http-bridge

remorhaz:
So, should I move my code to separate repository?

kelunik:
@remorhaz Do you want to maintain it yourself or should I create a new repository in the @amphp namespace? I'd be fine with both.

remorhaz:
I guess it's better to keep it in amphp space.

remorhaz:
@kelunik, so what about a new repository?

kelunik:
@remorhaz I've just created https://github.com/amphp/http-client-psr7, so you can PR against that. :-)

enumag:
ping @remorhaz

remorhaz:
Yep, @enumag, I remember; just havin' some hot time on main job. I'll make a PR in a day or two.

kelunik:
Closing, because of https://github.com/amphp/http-client-psr7/commit/487af2548c154c8a1d61bae2692817575b882e39.

PR #274, opened 2020-07-13
Add addFileFromString method.

Pato05:
Just a little function to add a file with a custom fileName and content-type

### DIFF
--- a/src/Body/FormBody.php
+++ b/src/Body/FormBody.php
@@ -149,7 +149,11 @@ final class FormBody implements RequestBody
             $fields[] = "--{$this->boundary}\r\n";
 
             /** @psalm-suppress PossiblyNullArgument */
+<<<<<<< HEAD
             $fields[] = $fileName !== null
+=======
+            $fields[] = $field instanceof FileBody
+>>>>>>> parent of 9899cbd... Add Content-Length header and make some improvements.
                 ? $this->generateMultipartFileHeader($name, $fileName, $contentType)
                 : $this->generateMultipartFieldHeader($name, $contentType);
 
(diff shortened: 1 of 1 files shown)


src/Body/FormBody.php
```diff
@@ -150,8 +165,7 @@ private function getMultipartFieldArray(): array
     private function generateMultipartFileHeader(string $name, string $fileName, string $contentType): string
     {
         $header = "Content-Disposition: form-data; name=\"{$name}\"; filename=\"{$fileName}\"\r\n";
-        $header .= "Content-Type: {$contentType}\r\n";
-        $header .= "Content-Transfer-Encoding: binary\r\n\r\n";
```
kelunik, line 154: Why did you remove this line?

src/Body/FormBody.php
```diff
@@ -164,6 +178,7 @@ private function generateMultipartFieldHeader(string $name, string $contentType)
         } else {
             $header .= "\r\n";
         }
+        $header .= "\r\n";
```
kelunik, line 181: Why did you add this line?

kelunik:
Merged manually, thanks!

src/Body/FormBody.php
Pato05: Because afaik Content-Transfer-Encoding is deprecated and about Content-Type I think I readded it?

src/Body/FormBody.php
Pato05: Just to make the code look nicer, I removed the carrier return on the upper parts

Pato05:
Ok, thank you :D

src/Body/FormBody.php
kelunik: Yeah, not sure whether this is still needed for anyone, but it doesn't really hurt, either.

src/Body/FormBody.php
Pato05: Oh yea, sorry, my fault xd

src/Body/FormBody.php
Pato05: Yeah, true... Indeed I kinda removed it to do some tests, then I saw that it was deprecated so I didn't rebring it there.

Issue #277, opened 2020-11-02
ConnectionLimitingPool closing connections while in use


descawed:
I've been troubleshooting some connection errors in my application and I've tracked them down to a couple issues in ConnectionLimitingPool:

- In `getStreamFor`, when a new connection is created, the connection ID is captured so it can be passed to `dropConnection` when the connection is closed. But what it's actually capturing is the ID of the connection promise, not the connection itself. This causes `dropConnection` to not correctly remove the connection from `activeRequestCounts` and `idleConnections`.
- The `onReadyConnection` method closes idle connections if it detects that there are too many open. But it does so with `array_shift`, which re-indexes the array starting from 0, so the remaining idle connections are no longer mapped to the correct connection ID.

These errors result in connections not being removed the list of idle connections when they should be, which can lead to connections being closed while still in use. I'll submit a PR for this momentarily.

kelunik:
You seem to be running into some pretty hard to debug edge cases. Thank you!

PR #278, opened 2020-11-02
Fix ConnectionLimitingPool closing non-idle connections

descawed:
Fix for #277. I've included a test as well, although it's kind of a mess. The closures modifying each other's values by reference is kind of hacky, and the value of $numRequests is based on the hard-coded limit of 64 in `ConnectionLimitingPool->onReadyConnection`.  I had to actually implement the onClose callbacks to test when the pool closes idle connections, and I don't think I can use a PHPUnit expectation to verify when the connection is closed because we need to know that the connection was closed after the request promise was fulfilled, not just after the method was called. I'm open to any suggestions on how to improve it.


kelunik:
Hey @descawed, thanks for the fix!

PR #284, opened 2020-12-23
request() needs Request as the parameter, not a string

ktecho:
$client->request now uses a Request as the parameter, not a url string


kelunik:
Thanks!

Issue #101, opened 2016-11-03
PSR-7 support


alexeyshockov:
Thanks for great feature-rich HTTP client implementation!

Do you have any plans for PSR-7 support? Standard request/response interfaces and so on.

bwoebi:
No.

In particular the Streams API of PSR-7 is not compatible with the Promise based API.

It should be easy though, to write a blocking wrapper implementing PSR-7 around Artax. I do not plan to write one, but users are welcome to write their own.

kelunik:
**Further Explanation**

PSR-7's `StreamInterface` doesn't work well in an async context without fully buffering all bodies, defeating the entire purpose of using streams. Major problems are `__toString()`, `getContents()`, etc.

[ReactPHP tried to implement PSR-7 in their HTTP server](https://github.com/reactphp/http/blob/801e8e5357210bcda5a1600dcf852860ccdc8886/src/HttpBodyStream.php), but it's not fully compliant. Most stream methods just throw a `BadMethodCallException`.

In our opinion it doesn't make sense to migrate to such a broken implementation if it can't be fully compatible without breaking basic async principles or the contract.

We have [our own interfaces for streams](https://github.com/amphp/byte-stream) that work well with async that are used across many of our packages.

Additionally, PSR-7 has several weaknesses such as defining streams and URIs inside an HTTP abstraction specification.

PR #299, opened 2022-01-31
Fix issues caused by new betas

iggyvolz:
Still have a few tests failing that I'm not quite sure how to fix but overall seems to be working after a few fixups:

- ClientHttpBinIntegrationTest::testHttp2Push `Amp\Socket\ConnectException: Connection to tcp://http2-server-push-demo.keksi.io:443 failed: [Error #101] Network is unreachable` (might be some issue with my network setup)
- ClientHttpBinIntegrationTest::testHttp2TeHeader `PHPUnit\Framework\AssertionFailedError: No response callback set`
- ClientHttpBinIntegrationTest::testFileBodyRequest, ClientHttpBinIntegrationTest::testMultipartBodyRequest, FormBodyTest::testMultiPartFields `Error: Call to undefined function Amp\Pipeline\fromIterable()` looks to be from here https://github.com/amphp/parallel/blob/1de63bae78f5bad50f31146370e015a65dc71295/src/Sync/ChannelledStream.php#L49 but updating that causes a dependency hell that I haven't been able to work out yet - might be easier to let this one sit for a bit
- ClientHttpBinIntegrationTest::testContentLengthBodyMismatch, Http1ConnectionTest::testWritingRequestWithValidUrlPathProceedsWithMatchingUriPath with data set "Empty path is replaced with slash" `Amp\ByteStream\StreamException thrown to event loop error handler: Failed to write to stream (8): fwrite(): Send of 1 bytes failed with errno=32 Broken pipe` - not sure what's going on here
-  LogIntoHttpArchiveTest::testProducesValidJson `Failed asserting that an empty string is valid JSON.`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -80,5 +80,23 @@
         "branch-alias": {
             "dev-master": "4.x-dev"
         }
+    },
+  "repositories": [
+    {
+      "type": "git",
+      "url": "https://github.com/iggyvolz/socket"
+    },
+    {
+      "type": "git",
+      "url": "https://github.com/iggyvolz/cache"
+    },
+    {
+      "type": "git",
+      "url": "https://github.com/iggyvolz/process"
+    },
+    {
+      "type": "git",
+      "url": "https://github.com/iggyvolz/dns"
     }
+  ]
 }
--- a/composer.json
+++ b/composer.json
@@ -80,23 +80,5 @@
         "branch-alias": {
             "dev-master": "4.x-dev"
         }
-    },
-  "repositories": [
-    {
-      "type": "git",
-      "url": "https://github.com/iggyvolz/socket"
-    },
-    {
-      "type": "git",
-      "url": "https://github.com/iggyvolz/cache"
-    },
-    {
-      "type": "git",
-      "url": "https://github.com/iggyvolz/process"
-    },
-    {
-      "type": "git",
-      "url": "https://github.com/iggyvolz/dns"
     }
-  ]
 }
--- a/src/Connection/Internal/Http2ConnectionProcessor.php
+++ b/src/Connection/Internal/Http2ConnectionProcessor.php
@@ -31,8 +31,8 @@ use Amp\Http\Http2\Http2Processor;
 use Amp\Http\Http2\Http2StreamException;
 use Amp\Http\InvalidHeaderException;
 use Amp\Http\Status;
-use Amp\Pipeline\Emitter;
 use Amp\Pipeline\Pipeline;
+use Amp\Pipeline\Queue;
 use Amp\Socket\EncryptableSocket;
 use Amp\TimeoutCancellation;
 use League\Uri;
@@ -103,7 +103,7 @@ final class Http2ConnectionProcessor implements Http2Processor
 
     private int|null $shutdown = null;
 
-    private Emitter $frameQueueSource;
+    private Queue $frameQueueSource;
 
     private Pipeline $frameQueue;
 
@@ -111,7 +111,7 @@ final class Http2ConnectionProcessor implements Http2Processor
     {
         $this->socket = $socket;
         $this->hpack = new HPack;
-        $this->frameQueueSource = new Emitter();
+        $this->frameQueueSource = new Queue();
         $this->frameQueue = $this->frameQueueSource->pipe();
     }
 
@@ -435,7 +435,7 @@ final class Http2ConnectionProcessor implements Http2Processor
             }
         });
 
-        $stream->body = new Emitter;
+        $stream->body = new Queue();
         $stream->trailers = new DeferredFuture;
         $stream->trailers->getFuture()->ignore();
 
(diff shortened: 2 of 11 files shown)


composer.json
```diff
@@ -80,5 +80,23 @@
         "branch-alias": {
             "dev-master": "4.x-dev"
         }
+    },
+  "repositories": [
```
iggyvolz: Temporary until amphp/dns#106, amphp/process#62, amphp/cache#19, amphp/socket#87

test/ResponseTest.php
```diff
@@ -88,9 +88,6 @@ public function testBody(): void
         $response->setBody("foobar");
         $this->assertSame('foobar', $response->getBody()->buffer());
 
-        $response->setBody($response->getBody());
```
iggyvolz, line 91: Test was throwing an error that you can't `->buffer()` twice, which seems to be exactly what's being tested here - guessing that behaviour was removed.

test/ClientHttpBinIntegrationTest.php
```diff
@@ -21,17 +21,18 @@
...
+use Amp\Http\Server\RequestHandler\ClosureRequestHandler;
 use Amp\Http\Server\Response as ServerResponse;
 use Amp\Http\Server\Server;
 use Amp\PHPUnit\AsyncTestCase;
+use Amp\Pipeline\Pipeline;
 use Amp\Socket;
+use Closure;
 use Psr\Log\NullLogger;
 use Revolt\EventLoop;
+use function _HumbugBoxae3c412aa099\Amp\asyncCall;
```
trowski: Oops?

test/ClientHttpBinIntegrationTest.php
```diff
@@ -588,7 +589,7 @@ public function getHeaders(): array
 
             public function createBodyStream(): ReadableStream
             {
-                return new IterableStream(fromIterable(["a", "b", "c"])->pipe(postpone(500)));
+                return new IterableStream(Pipeline::fromIterable(["a", "b", "c"])->map(fn() => delay(500)));
```
trowski: This should be `Pipeline::delay(500)` instead.

test/Connection/Http1ConnectionTest.php
```diff
@@ -259,8 +259,8 @@ public function getHeaders(): array
 
             public function createBodyStream(): ReadableStream
             {
-                $pipeline = Pipeline\fromIterable(\array_fill(0, 100, '.'));
-                $pipeline = $pipeline->pipe(Pipeline\postpone(0.1));
+                $pipeline = Pipeline::fromIterable(\array_fill(0, 100, '.'));
+                $pipeline = $pipeline->map(fn() => delay(0.1));
```
trowski: Same here.

test/ClientHttpBinIntegrationTest.php
iggyvolz: Gah PHPStorm tried to autocomplete that...

src/Connection/Internal/Http2ConnectionProcessor.php
```diff
@@ -1756,9 +1756,7 @@ private function createStreamInactivityWatcher(int $streamId, float $timeout): ?
     private function runWriteThread(): void
     {
         try {
-            while (null !== $frame = $this->frameQueue->continue()) {
-                $this->socket->write($frame);
-            }
+            $this->frameQueue->forEach(fn($frame) => $this->socket->write($frame));
```
kelunik, line 1759: Instead of using `Pipeline`, we can use `ConcurrentIterator` directly here.

src/Connection/Internal/Http2ConnectionProcessor.php
```diff
@@ -103,15 +103,15 @@ final class Http2ConnectionProcessor implements Http2Processor
...
 
     private Pipeline $frameQueue;
 
     public function __construct(EncryptableSocket $socket)
     {
         $this->socket = $socket;
         $this->hpack = new HPack;
-        $this->frameQueueSource = new Emitter();
+        $this->frameQueueSource = new Queue();
         $this->frameQueue = $this->frameQueueSource->pipe();
```
kelunik, line 115: It probably makes sense to rename `frameQueueSource` to `frameQueue` and `frameQueue` to `frameQueueIterator`

test/ClientHttpBinIntegrationTest.php
```diff
@@ -21,17 +21,17 @@
...
 use Amp\Http\Server\Options;
 use Amp\Http\Server\Request as ServerRequest;
-use Amp\Http\Server\RequestHandler\CallableRequestHandler;
+use Amp\Http\Server\RequestHandler\ClosureRequestHandler;
 use Amp\Http\Server\Response as ServerResponse;
 use Amp\Http\Server\Server;
 use Amp\PHPUnit\AsyncTestCase;
+use Amp\Pipeline\Pipeline;
 use Amp\Socket;
+use Closure;
```
kelunik, line 30: We usually don't import global symbols but directly reference them using `\Closure` instead.

test/ResponseTest.php
kelunik: Yes, that has been removed in https://github.com/amphp/byte-stream/commit/2c3f3383bc7587ae1997ad3da40486e83351ff0d#diff-ebb3b05d8384827669dfc2fadbce197f021879aed9b438e97e931bb498e0cef2. The issues I see with offering it again is different cancellations / limits on multiple calls.

@trowski This has been a hidden change, but I think it was the correct thing to do, no?

kelunik:
Merged this already, so I can continue working on it.

PR #308, opened 2022-11-08
Remove left-over comment from v4

stof:
There is no promise anymore in the signature.


kelunik:
Thanks! There's no promise anymore, but the requirement of the first stream request providing a valid stream still stays.

kelunik:
Done in https://github.com/amphp/http-client/commit/276025749f893cfc221538141a8d04f2102a33e5.

PR #322, opened 2023-04-10
Rework body API

kelunik:


src/Connection/Http1Connection.php
```diff
@@ -297,9 +297,8 @@ private function readResponse(
                 $status = $response->getStatus();
 
                 if ($status === Http\HttpStatus::SWITCHING_PROTOCOLS) {
-                    $connection = Http\parseMultipleHeaderFields($response, 'connection')[0] ?? null;
-
-                    if (!isset($connection['upgrade'])) {
+                    $connection = array_map('strtolower', Http\splitHeader($response, 'connection'));
+                    if (!\in_array('upgrade', $connection, true)) {
```
kelunik: @trowski We should probably make this simpler with an API provided by `amphp/http`.

trowski: Hmm, yes I didn't think about header casing. Perhaps we want to always lowercase the keys in `parseMultipleHeaderFields`? Then I think the code would look like

```php
$connection = Http\parseMultipleHeaderFields($response, 'connection')[0] ?? null;
if (isset($connection['upgrade']) {
    // ...
}
```

rekryt: How about
```php
$connection = Http\parseMultipleHeaderFields($response, 'connection')[0] ?? [];
if (!\array_key_exists('upgrade', $connection)) {
    // ...
}
```

kelunik: @rekryt The issue here is the casing, I think.

trowski: It might be both actually. Maybe keys without a value should have an empty string as the value.

rekryt: Keys without a value should have an "null" as value.
```
$connection = [ 'upgrade' => null ];
isset($connection['upgrade']) === false;
```

kelunik: Updated to the newly introduced `parseHeaderTokens`.

Bilge:
>No description provided.

lol thanks

PR #323, opened 2023-04-10
Header "upgrade" existence check on Http\HttpStatus::SWITCHING_PROTOCOLS

rekryt:
```
$connection = [ 'upgrade' => null ];
echo isset($connection['upgrade']); // false
```


kelunik:
Thanks! I've already included a newer version of the fix in #322.

PR #328, opened 2023-05-11
allow also psr/http-message 2.*

drealecs:
allow also psr/http-message 2.*

There is no backward incompatibility for v4 as psr/http-message for 2.0 has also PHP 7.2 as minimum requirement.

The branch 4.x can be merged into 5.x as that's valid for 5.x as well and it's not yet changed there.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
         "amphp/socket": "^1",
         "amphp/sync": "^1.3",
         "league/uri": "^6",
-        "psr/http-message": "^1"
+        "psr/http-message": "^1 | ^2"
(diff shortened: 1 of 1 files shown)


kelunik (approved):

kelunik:
Thanks, @drealecs!

GrahamCampbell:
This change doesn't do anything because of the league uri dependency. I think it should be removed from the change log, since it's super misleading.

kelunik:
Good point, I've removed it from the changelog, as this PR is for 4.x anyway and not 5.x.

PR #330, opened 2023-05-16
Ensure HttpStream $releaseCallback is called on destruct

### LINKED ISSUE #329: Memory issue in HTTP 2 connection
Related to #324, while the reproducer in that issue was indeed fixed, we still experienced the issue on production.

I have narrowed the issue down to HTTP 2 connections, since httpbin.org seems to not support HTTP 2, the reproducer in #324 was fixed. Note that due to some internal reasons, we are not able to re-use the HttpClient for long and as such must re-create it periodically which causes the problem.

New reproducer uses an alternative httpbin that supports HTTP 2:
```
<?php

declare(strict_types=1);

use Amp\Http\Client\Request;
use function Amp\async;

include __DIR__ . '/../vendor/autoload.php';

function reportMemory(): void
{
    gc_collect_cycles();
    gc_mem_caches();
    echo memory_get_usage() . PHP_EOL;
}

reportMemory();
for ($i = 0; $i < 10; ++$i) {
    async(static function () {
        $client = (new HttpClientBuilder())->build();
        $request = new Request('https://www.nghttp2.org/httpbin/');
        $request->setProtocolVersions(['2']);
        $client->request($request)->getBody()->buffer();
    })->await();

    reportMemory();
}
```

Output:
```
1385408
37151456
37225792
37304112
37381120
37458128
...
```

I narrowed the problem down to the `HttpStream::$releaseCallback` not being called, because it is set to `null` in its `request(...)` method. This in turn prevents the connection from reaching an idle state and prevents clean up. I'll submit a pull request shortly.

FlorianVen:
Fixes #329

### DIFF
--- a/src/Connection/Internal/Http2ConnectionProcessor.php
+++ b/src/Connection/Internal/Http2ConnectionProcessor.php
@@ -1326,12 +1326,6 @@ final class Http2ConnectionProcessor implements Http2Processor
 
         unset($this->streams[$streamId]);
 
-        if ($streamId & 1) { // Client-initiated stream.
-            $this->remainingStreams++;
-
-            \assert($this->remainingStreams <= $this->concurrentStreamLimit);
-        }
-
         if ($stream->responsePending || $stream->body || $stream->trailers) {
             $exception ??= new SocketException(
                 \sprintf("Stream %d closed unexpectedly", $streamId),
--- a/src/Connection/HttpStream.php
+++ b/src/Connection/HttpStream.php
@@ -52,6 +52,8 @@ final class HttpStream implements Stream
     /** @var callable|null */
     private $releaseCallback;
 
+    private bool $used = false;
+
     private function __construct(
         SocketAddress $localAddress,
         SocketAddress $remoteAddress,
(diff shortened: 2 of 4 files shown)


src/Connection/Internal/Http2ConnectionProcessor.php
```diff
@@ -1326,12 +1326,6 @@ private function releaseStream(int $streamId, ?\Throwable $exception = null): vo
 
         unset($this->streams[$streamId]);
 
-        if ($streamId & 1) { // Client-initiated stream.
-            $this->remainingStreams++;
-
-            \assert($this->remainingStreams <= $this->concurrentStreamLimit);
-        }
-
```
kelunik, line 1334: Why did you remove these lines?

FlorianVen: Because the streams $releaseCallback is set to Http2Processor::unreserveStream (Http2Connection line 60) which also increments $remainingStreams, subsequently the assertion would be false because $remainingStreams was incremented twice but only decremented once.

Either reserveStream and unreserveStream should *only* influence $reservedStreams, or they should be the only ones to influence both. I choose the later because I couldn't find a good place to decrement $remainingStreams elsewhere in Http2ConnectionProcessor

FlorianVen: @kelunik Checking in, does the above seem reasonable?

trowski: I thought this was strange when I first looked at the PR, but yes, this seems correct if instead the unreserveStream callback is invoked upon stream release.

kelunik:
Fixed via https://github.com/amphp/http-client/commit/d92c290fd30c21e7a93f0af0da38c55db1ee42aa instead.

PR #334, opened 2023-07-02
New event system

kelunik:


src/Connection/Internal/Http1Parser.php
```diff
@@ -116,9 +116,9 @@ public function parse(string $data = null): ?Response
         if (!$this->bodyStarted && \in_array($this->state, [self::BODY_CHUNKS, self::BODY_IDENTITY, self::BODY_IDENTITY_EOF], true)) {
             $this->bodyStarted = true;
 
+            $eventInvoker = events();
             \assert($this->response !== null);
-
-            events()->responseBodyStart($this->request, $this->stream, $this->response);
+            $eventInvoker->responseBodyStart($this->request, $this->stream, $this->response);
```
kelunik: Why this change?

trowski: The Psalm setting `rememberPropertyAssignmentsAfterCall="false"` means calling `events` loses the assertion that `$this->response !== null`

kelunik: Shouldn't we store `$this->response` in a variable then instead of the event invoker?

composer-require-check.json
```diff
@@ -16,6 +16,9 @@
         "void",
         "object",
         "mixed",
+        "random_bytes",
+        "events",
+        "processRequest",
```
kelunik: Those are double present now, see below.

trowski: Oops, automerge duplicated that since we put them in different places. Fixed.

src/Connection/Internal/Http1Parser.php
trowski: Either way, but that does match the style elsewhere better. Updated.

src/Connection/Internal/Http2ConnectionProcessor.php
```diff
@@ -732,7 +717,7 @@ public function handleStreamException(Http2StreamException $exception): void
         $exception = new SocketException($exception->getMessage(), $code, $exception);
 
         if ($code === Http2Parser::REFUSED_STREAM) {
-            $exception = new UnprocessedRequestException($exception);
+            $exception = new UnprocessedRequestException($exception); // TODO
```
trowski: There's still a TODO here, something missing yet?

src/Connection/Internal/Http2ConnectionProcessor.php
```diff
@@ -1441,7 +1413,7 @@ private function shutdown(?HttpException $reason = null, ?int $lastId = null): v
             if ($lastId !== null && $id > $lastId) {
                 $exception = $exception instanceof UnprocessedRequestException
                     ? $exception
-                    : new UnprocessedRequestException($reason);
+                    : new UnprocessedRequestException($reason); // TODO
```
trowski: Same here.

src/Connection/Internal/Http2ConnectionProcessor.php
```diff
@@ -1035,19 +1017,18 @@ public function request(Request $request, Cancellation $cancellation, Stream $st
                     $writeFuture->await($cancellation);
 
                     if ($chunk === null) {
-                        $http2stream->requestBodyCompletion->complete();
+                        $http2stream->requestBodyCompletion->complete(); // TODO Move after loop?
```
trowski: I think moving this after loop would be better to be triggered at a similar time to the event listener.

src/Connection/Internal/Http2ConnectionProcessor.php
kelunik: The branch gets rid of `UnprocessedRequestException` mostly, but not entirely, yet. Not sure why I mixed those changes into one branch.

PR #335, opened 2023-07-03
Allow league/uri ^7

nicolas-grekas:
So that psr/http-message 2.* can be installed (see #328)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
         "amphp/http": "^1.6",
         "amphp/socket": "^1",
         "amphp/sync": "^1.3",
-        "league/uri": "^6",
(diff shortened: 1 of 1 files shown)


kelunik:
Let's wait for the stable release. 🙂

kelunik:
@nicolas-grekas Could you please re-trigger the pipeline?

nicolas-grekas:
force-pushed but still failing

kelunik:
@nicolas-grekas Thanks! We've fixed the breaking change in hpack in https://github.com/amphp/hpack/releases/tag/v3.1.3. There's one remaining test failure, but that seems unrelated.

PR #339, opened 2023-08-01
Fixed form content encoding

### LINKED ISSUE #333: `Form` not sent with same semantics as `FormBody`
I apologise I haven't been able to reduce this down to a simpler test case, but it seems very clear to me that something is not right when sending `Form` in beta 11 as with `FormBody` from the previous API iteration. I haven't been able to narrow down exactly what is different between the two requests, but I can reliably reproduce it with a test.

To test this, please try running [`SteamLoginTest::testSecureLoginCookie`](https://github.com/Provider/Steam/blob/master/test/Functional/SteamLoginTest.php), which requires `STEAM_USER` and `STEAM_PASSWORD` to be set to a valid Steam username and password (you can [get an account](https://store.steampowered.com/join) for free).

`master` will work, but switching over to `connectors-v7`, which has [minimal changes](https://github.com/Provider/Steam/compare/connectors-v7), will fail on `SteamLogin:99` with `{"response":{"interval":5,"extended_error_message":""}}`. Unfortunately this opaque error is all we have to work with.

Bilge:
Changed form content encoding from `PHP_QUERY_RFC3986` -> `PHP_QUERY_RFC1738`. This was the default in v5.0.0-beta.7.

It seems this inadvertently changed when [http_build_query()](https://github.com/amphp/http-client/blob/v5.0.0-beta.7/src/Body/FormBody.php#L192) was replaced with `League\Uri\QueryString` for whatever reason. The [default](https://github.com/thephpleague/uri-components/blob/06f49e726a7fd06063e80f5f8b045fb9128e6c43/src/QueryString.php#L257) in that library is different.

Closes #333.

### DIFF
--- a/src/Form.php
+++ b/src/Form.php
@@ -165,7 +165,7 @@ final class Form implements HttpContent
             }
         }
 
-        return QueryString::build($pairs) ?? '';
+        return QueryString::build($pairs, enc_type: PHP_QUERY_RFC1738) ?? '';
     }
 
     /**


kelunik:
Thanks, manually merged as https://github.com/amphp/http-client/commit/4a962fe1d96ca0ca956401f08eb3eb72e90e5d2b with a test.

Bilge:
Why didn't you use my commit?

kelunik:
Because I wrote a failing test locally first, then fixed it and then the test was green, but I already had everything ready to commit locally, so I just committed. I also wanted to avoid using named params for libraries we don't control.

PR #342, opened 2023-08-12
Draft: Fix request cloning for double redirects

kelunik:


src/Interceptor/FollowRedirects.php
```diff
@@ -188,21 +188,22 @@ private function createRedirectRequest(Request $clonedRequest, Response $respons
...
-        $clonedRequest->setUri($redirectUri);
+        $request = $this->cloneRequest($clonedRequest);
+        $request->setUri($redirectUri);
 
         if (!$isSameHost) {
             // Avoid copying headers for security reasons, any interceptor headers will be added again,
             // but application headers will be discarded.
-            $clonedRequest->setHeaders([]);
+            $request->setHeaders([]);
         }
```
kelunik, line 198: @trowski We assign the original headers here if `https://example.com` redirects to `https://example.net` and then back `https://example.com`, but this seems to be mostly safe?

kelunik: Nevermind, this is actually problematic, because it keeps the headers if request 2 and request 3 are the same host.

PR #343, opened 2023-08-14
Fixed multiple redirects by removing addAcceptEncodingHeader()

### LINKED ISSUE #341: Double redirect still returns gzip compressed data
Non-redirect responses are decompressed implicitly, redirected (once) responses are decompressed implicitly, but double-redirected responses are _not_ decompressed implicitly and return the raw gzip data.

Bilge:
Fixes #341.


kelunik:
This just disables compression, it fixes the symptom, but not the root cause, which is in FollowRedirects.

PR #355, opened 2023-12-10
Add JsonContent

trowski:
Adds a simple `JsonContent` request body implementation, as sending a JSON body is very common.


kelunik (commented):
I think this should be a function reusing a generic implementation under the hood.

trowski:
@kelunik I made it a static method on `BufferedContent`. Does this seem appropriate or do you think it should be a regular function? If so, do you have a name suggestion? I was thinking `createJsonContent`.

kelunik (approved):

PR #356, opened 2023-12-10
Add addNestedFields() to Form

trowski:
Adds a method to `Form` to insert an array of values into the form, preserving array keys.

### DIFF
--- a/src/Form.php
+++ b/src/Form.php
@@ -52,13 +52,34 @@ final class Form implements HttpContent
     /**
      * Adds each member of the array as an entry for the given key name. Array keys are persevered.
      *
-     * @param array<string> $fields
+     * @param array<string|array> $fields
      */
     public function addNestedFields(string $name, array $fields): void
     {
-        foreach ($fields as $key => $content) {
-            $this->addField(\sprintf('%s[%s]', $name, $key), $content);
+        foreach ($this->flattenArray($fields) as $key => $value) {
+            $this->addField($name . $key, $value);
         }
+    }
+
+    /**
+     * @return array<string, string>
+     */
+    private function flattenArray(array $fields): array
+    {
+        $result = [];
+        foreach ($fields as $outerKey => $value) {
+            $key = "[{$outerKey}]";
+            if (!\is_array($value)) {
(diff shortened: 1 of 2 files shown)


src/Form.php
```diff
@@ -49,6 +49,18 @@ public function addField(string $name, string $content, ?string $contentType = n
...
 
+    /**
+     * Adds each member of the array as an entry for the given key name. Array keys are persevered.
+     *
+     * @param array<string> $fields
+     */
+    public function addFields(string $name, array $fields): void
+    {
+        foreach ($fields as $key => $content) {
+            $this->addField(\sprintf('%s[%s]', $name, $key), $content);
```
kelunik: This is PHP specific syntax, no?

src/Form.php
```diff
@@ -49,6 +49,18 @@ public function addField(string $name, string $content, ?string $contentType = n
         $this->fields[] = new FormField($name, BufferedContent::fromString($content, $contentType));
     }
 
+    /**
+     * Adds each member of the array as an entry for the given key name. Array keys are persevered.
+     *
+     * @param array<string> $fields
+     */
+    public function addFields(string $name, array $fields): void
```
kelunik: Based on our usual naming, I'd expect this to add a repeated field, not nested fields based on the array keys.

src/Form.php
trowski: No, it seems this is a de-facto "standard" for arrays in query strings. I found several libraries supporting it. Here's one for example in python: https://github.com/bernii/querystring-parser

src/Form.php
trowski: I agree. `addArrayField`? Or something else?

kelunik: Maybe `addNestedFields`?

trowski: Works for me. Updated.

src/Form.php
kelunik: Should it support any depth then?

trowski: Sure, probably uncommon, but that makes sense. Updated.

kelunik (approved):

PR #357, opened 2023-12-13
add unix socket support

ahjdev:


kelunik:
Unix sockets are already supported, see https://github.com/amphp/http-client/blob/5.x/examples%2Fbasic%2F5-unix-sockets.php

PR #362, opened 2024-04-26
Add inactivity timeout to SetRequestTimeout interceptor

Nek-:
I figured out that the inactivity timeout was not present in the default interceptor in charge of updating the timeouts.

Of course it's not a big deal and here is a workaround:

```php
$this->httpClient = (new HttpClientBuilder)
    ->interceptNetwork(
        new ModifyRequest(static function (Request $request) {
            $request->setInactivityTimeout(500);

            return $request;
        })
    )
    ->build();
```

But I decided maybe it's a good first contribution here 😄 . I also updated a bit the documentation because it was missing in the interceptor list! However, I'm sorry I don't know how to fix the test.

### DIFF
--- a/src/Interceptor/SetRequestTimeout.php
+++ b/src/Interceptor/SetRequestTimeout.php
@@ -10,15 +10,21 @@ final class SetRequestTimeout extends ModifyRequest
         float $tcpConnectTimeout = 10,
         float $tlsHandshakeTimeout = 10,
         float $transferTimeout = 10,
+        float $inactivityTimeout = null,
     ) {
         parent::__construct(static function (Request $request) use (
             $tcpConnectTimeout,
             $tlsHandshakeTimeout,
-            $transferTimeout
+            $transferTimeout,
+            $inactivityTimeout
         ) {
             $request->setTcpConnectTimeout($tcpConnectTimeout);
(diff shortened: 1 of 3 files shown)


trowski:
Thanks! It looks like the test you added is failing due to a mis-matched message. Could you take a look at that?

Nek-:
Oops. Sorry, it was late! 😂 

Fixed it! 🙏

src/Interceptor/SetRequestTimeout.php
```diff
@@ -10,15 +10,18 @@ public function __construct(
         float $tcpConnectTimeout = 10,
         float $tlsHandshakeTimeout = 10,
         float $transferTimeout = 10,
+        float $inactivityTimeout = 10,
```
kelunik: Should this default to `null` to prevent modifying the inactivity timeout if there's already a custom timeout present?

Nek-: You mean for BC?

kelunik: Yes

Nek-: Changed!

kelunik (approved):

kelunik:
Thanks!

PR #370, opened 2024-12-09
Uri\Http deprecated

aleswita:
From version `7.5` of `league/uri` got deprecation warning: `E_USER_DEPRECATED: Method League\Uri\Http::createFromString() is deprecated since league/uri:7.0.0, use League\Uri\Http::new() instead`

This code shoud be compatible with `league/uri` `^6|^7`


xximwon (approved):

trowski:
Merged a more complete fix in #372. Thank you though! :+1:

PR #378, opened 2025-02-20
Test against PHP 8.4

Nek-:
I already use the http client with PHP 8.4. But I figured out that the devkit was limited to PHP 8.3. Let's update things!

Outch. It has been a battle to update psalm configuration. I added some new rules to avoid many many MANY useless casts.

Wont be merged, see https://github.com/amphp/http-client/pull/378#issuecomment-2690292872

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,7 +48,7 @@
         "phpunit/phpunit": "^9",
         "amphp/http-server": "^3",
         "kelunik/link-header-rfc5988": "^1",
-        "psalm/phar": "~5.23",
+        "psalm/phar": "*",
         "laminas/laminas-diactoros": "^3.5.0"
     },
     "suggest": {
--- a/composer.json
+++ b/composer.json
@@ -49,7 +49,7 @@
         "amphp/http-server": "^3",
         "kelunik/link-header-rfc5988": "^1",
         "psalm/phar": "~5.23",
-        "laminas/laminas-diactoros": "^2.3"
+        "laminas/laminas-diactoros": "^3.5.0"
     },
     "suggest": {
         "ext-zlib": "Allows using compression for response bodies.",
--- a/psalm.xml
+++ b/psalm.xml
@@ -4,9 +4,12 @@
         phpVersion="8.1"
         resolveFromConfigFile="true"
         rememberPropertyAssignmentsAfterCall="false"
+        ensureOverrideAttribute="false"
+        findUnusedCode="false"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xmlns="https://getpsalm.org/schema/config"
         xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
+        xmlns:xi="http://www.w3.org/2001/XInclude"
 >
     <projectFiles>
         <directory name="examples"/>
@@ -23,6 +26,20 @@
                 <directory name="src"/>
             </errorLevel>
         </StringIncrement>
+
+        <ImplicitToStringCast>
+            <errorLevel type="suppress">
+                <directory name="examples"/>
+                <directory name="src"/>
+            </errorLevel>
+        </ImplicitToStringCast>
+
+        <InvalidOperand>
+            <errorLevel type="suppress">
+                <directory name="examples"/>
+                <directory name="src"/>
+            </errorLevel>
+        </InvalidOperand>
 
         <DuplicateClass>
             <errorLevel type="suppress">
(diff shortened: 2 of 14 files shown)


composer.json
```diff
@@ -48,8 +48,8 @@
         "phpunit/phpunit": "^9",
         "amphp/http-server": "^3",
         "kelunik/link-header-rfc5988": "^1",
-        "psalm/phar": "~5.23",
-        "laminas/laminas-diactoros": "^2.3"
+        "psalm/phar": "*",
```
Nek-, line 51: The version will be 6 but different regarding the PHP version.

danog: You should at least require the latest `^6.8.5`

psalm.xml
```diff
@@ -4,9 +4,12 @@
         phpVersion="8.1"
         resolveFromConfigFile="true"
         rememberPropertyAssignmentsAfterCall="false"
+        ensureOverrideAttribute="false"
```
danog, line 7: Please set this to true, override attributes are recommended adn compatible with all PHP versions if the symfony/polyfill-php83 packages required

psalm.xml
```diff
@@ -4,9 +4,12 @@
...
+        findUnusedCode="false"
```
danog, line 8: Please keep this on

psalm.xml
```diff
@@ -59,5 +70,11 @@
                 <directory name="src" />
             </errorLevel>
         </RiskyTruthyFalsyComparison>
+
+        <InvalidReturnType>
```
danog, line 74: Don't suppress this at the file level

psalm.xml
```diff
@@ -24,6 +27,20 @@
...
         </StringIncrement>
 
+        <ImplicitToStringCast>
+            <errorLevel type="suppress">
+                <directory name="examples"/>
+                <directory name="src"/>
+            </errorLevel>
+        </ImplicitToStringCast>
+
+        <InvalidOperand>
```
danog, line 37: Don't suppress this at a project level

danog: https://github.com/vimeo/psalm/pull/11329 will fix what you're (incorrectly) trying to fix

psalm.xml
Nek-: It's a library, everything is unused code!

danog: No, public code must be marked with `@api`, the remaining code is private, unused code.

bwoebi: @danog I disagree - in a library everything is api, unless it's marked @internal, or is a private method.
This is littering the code with unnecessary annotations.
This brings no value as is, as I'm going to be only interested in whether my internal code is dead.

bwoebi: Please keep this config off unless psalm offers a way to "default-mark" everything non-internal & non-private as API.

danog: The problem with default-marking everything non-internal and non-private as API is with non-library codebases (that have no public surface apart from API controllers and other entry points).

Some not so clean library codebases have the same problem as well, where internal details are needlessly exposed as public and not marked as internal.

Explicitly marking public API with `@api` solves both issues.

psalm.xml
Nek-: It's not only that. The code often do some `float * int`, which is perfectly fine, btw `string . float` is also ok.

danog: If you have issues with that, just set strictBinaryOperands="false", do not suppress the issue as it also covers valid errors

psalm.xml
bwoebi: @danog I think it would be a good thing to make that configurable in psalm. `libraryAssumeApi="true"` or such.

composer.json
Nek-: Did you read the message I wrote just above?

psalm.xml
Nek-: We need lower versions of psalm for php versions compatibility. It would not work anyway.

composer.json
danog: Yes, and your assumption is wrong: you should always require the latest available version of Psalm.

psalm.xml
danog: Psalm is capable of running scans for versions down to php 5.4 on php 8.4, just pass the correct PHP version using the `--php-version` parameter.

danog: Psalm scans should **always** run on PHP 8.4.4 or above to avoid platform bugs, to target lower versions of PHP simply use the `--php-version` parameter.

composer.json
Nek-: The composer.json of psalm/phar does not allow this.

danog: It does, just require psalm 6.8.5

kelunik: We should definitely not require any `*` version without an upper bound. We should likely fix the psalm version, because newer versions often make builds fail due to new issues being detected.

Nek-:
Closing because it's complicated to make this PR as a person extern of the project.

Nice base to work if you need to use PHP 8.4 with this library in dev mode.

PR #382, opened 2025-03-04
Throws TimeoutException when TransferTimeout is exceeded

### LINKED ISSUE #379: SocketException distinction
There is an [Amp\Http\Client\SocketException](https://github.com/amphp/http-client/blob/5.x/src/SocketException.php) exception that covers different types of connection errors. However, I found out that it's not possible to distinguish between them programmatically, they are all the same except for the error message.
I need to differentiate a [specific error](https://github.com/amphp/http-client/blob/a3e8711cb71fe909c1ae17450bfa5db652559c20/src/Connection/Http1Connection.php#L454), and the only way to do that is to check if the error message contains a substring, which is not reliable.
What do you think we can do about this? My first thought is to add unique exception codes to all SocketException cases.

luzrain:
I found out that when setting a TransferTimeout limit on a request, the thrown exception is a `SocketException` instead of a `TimeoutException`. This issue only affects HTTP connections, HTTP/2 connections correctly throw `TimeoutException` in such cases as expected.
This change makes the code throw a `TimeoutException` instead of a `SocketException` when a connection session exceeds the TransferTimeout period.

Closes https://github.com/amphp/http-client/issues/379

### DIFF
--- a/src/Connection/Http1Connection.php
+++ b/src/Connection/Http1Connection.php
@@ -450,6 +450,7 @@ final class Http1Connection implements Connection
             }
 
             $originalCancellation->throwIfRequested();
+            $readingCancellation->throwIfRequested();
 
             throw new SocketException(\sprintf(
                 "Receiving the response headers for '%s' failed, because the socket to '%s' @ '%s' closed early with %d bytes received within %0.3f seconds",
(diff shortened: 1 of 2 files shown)


luzrain:
Here is the code to reproduce the issue.

```php
<?php

use Amp\Http\Client\Request;

require dirname(__DIR__, 1) . '/vendor/autoload.php';

$client = (new HttpClientBuilder())->retry(0)->build();

$request = new Request('https://httpbin.org/delay/5');
$request->setTransferTimeout(1);

try {
    $request1 = clone $request;
    $request1->setProtocolVersions(['1.1']);
    $response = $client->request($request1);
} catch (\Throwable $e) {
    var_dump('HTTP/1.1 ' . $e::class . ': ' . $e->getMessage());
}

try {
    $request2 = clone $request;
    $request2->setProtocolVersions(['2']);
    $response = $client->request($request2);
} catch (\Throwable $e) {
    var_dump('HTTP/2 ' . $e::class . ': ' . $e->getMessage());
}
```

The output without the fix:
Different exceptions are thrown in the same scenario for HTTP/1.1 and HTTP/2 requests
```
string(214) "HTTP/1.1 Amp\Http\Client\SocketException: Receiving the response headers for 'https://httpbin.org/delay/5' failed, because the socket to 'httpbin.org' @ '???' closed early with 0 bytes received within 1.000 seconds"
string(96) "HTTP/2 Amp\Http\Client\TimeoutException: Allowed transfer timeout exceeded, took longer than 1 s"
```

With this fix:
Exceptions are consistent for HTTP/1.1 and HTTP/2 requests
```
string(98) "HTTP/1.1 Amp\Http\Client\TimeoutException: Allowed transfer timeout exceeded, took longer than 1 s"
string(96) "HTTP/2 Amp\Http\Client\TimeoutException: Allowed transfer timeout exceeded, took longer than 1 s"
```

kelunik:
LGTM on first sight, but only looked on mobile.

trowski:
Thank you again!

PR #384, opened 2025-03-05
Proposal: TlsException to replace SocketException for TLS handshake errors

luzrain:
This PR proposes a new TlsException to replace SocketException in cases where an exception occurs during the TLS handshake.

Related: https://github.com/amphp/http-client/issues/379

### DIFF
--- a/src/SocketException.php
+++ b/src/SocketException.php
@@ -2,6 +2,6 @@
 
 namespace Amp\Http\Client;
 
-final class SocketException extends HttpException
+class SocketException extends HttpException
 {
 }
--- a/src/TlsException.php
+++ b/src/TlsException.php
@@ -1,0 +1,7 @@
+<?php declare(strict_types=1);
+
+namespace Amp\Http\Client;
+
+final class TlsException extends HttpException
+{
+}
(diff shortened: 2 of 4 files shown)


src/TlsException.php
```diff
@@ -0,0 +1,7 @@
+<?php declare(strict_types=1);
+
+namespace Amp\Http\Client;
+
+final class TlsException extends HttpException
```
kelunik: This need to extend SocketException to keep it backwards compatible.

kelunik (changes requested):

src/TlsException.php
luzrain: Done. I've moved final declaration in SocketException to dockblock, is it okay?

src/SocketException.php
```diff
@@ -2,6 +2,9 @@
 
 namespace Amp\Http\Client;
 
-final class SocketException extends HttpException
+/**
+ * @final
```
kelunik: We can skip the final if we extend it by design.

kelunik (changes requested):

src/SocketException.php
luzrain: ✔️

kelunik (approved):

kelunik:
Thank you again!

Issue #390, opened 2025-04-22
Connection leak when using HTTP/1


danog:
When using HTTP/1, connections are never released, leading to a pileup in the connection pool.

This is caused by the fact that the trailersDeferred created in Http1Connection::readResponse is never resolved.

Reproducer:
```
#!/usr/bin/env php
<?php declare(strict_types=1);

require dirname(__DIR__) . "/vendor/autoload.php";

use Amp\ByteStream;
use Amp\Http\Client\HttpClientBuilder;
use Amp\Http\Client\Request as ClientRequest;
use Amp\Http\HttpStatus;
use Amp\Http\Server\DefaultErrorHandler;
use Amp\Http\Server\Driver\SocketClientFactory;
use Amp\Http\Server\Request;
use Amp\Http\Server\RequestHandler;
use Amp\Http\Server\Response;
use Amp\Http\Server\SocketHttpServer;
use Amp\Log\ConsoleFormatter;
use Amp\Log\StreamHandler;
use Amp\Socket;
use Monolog\Logger;
use Monolog\Processor\PsrLogMessageProcessor;

use function Amp\async;
use function Amp\Future\await;
use function Amp\trapSignal;

$logHandler = new StreamHandler(ByteStream\getStdout());
$logHandler->pushProcessor(new PsrLogMessageProcessor());
$logHandler->setFormatter(new ConsoleFormatter());
$logger = new Logger('server');
$logger->pushHandler($logHandler);
$logger->useLoggingLoopDetection(false);

$server = new SocketHttpServer(
    $logger,
    new Socket\ResourceServerSocketFactory(),
    new SocketClientFactory($logger),
);

$server->expose("0.0.0.0:1337");
$server->expose("[::]:1337");

$server->start(new class implements RequestHandler {
    public function handleRequest(Request $request): Response
    {
        return new Response(
            status: HttpStatus::OK,
            headers: ["content-type" => "text/plain; charset=utf-8"],
            body: "Hello, World!",
        );
    }
}, new DefaultErrorHandler());

$client = HttpClientBuilder::buildDefault();
for ($x = 0; $x < 10; $x++) {
    $f = [];
    for ($y = 0; $y < 100; $y++) {
        $f []= async(function () use ($client) {
            $client->request(new ClientRequest("http://127.0.0.1:1337"));
        });
    }
    await($f);
}

// Await a termination signal to be received.
[...]

PR #391, opened 2025-04-23
Avoid leaking connections

danog:
Fix for https://github.com/amphp/http-client/issues/390.

Currently a WIP, as the current solution actually causes the destruction of the body reading fiber when the request is destroyed, instead of following through the proper body cancellation (or consumption) logic.


trowski:
Fix in https://github.com/amphp/http-client/commit/cf885bf00550d645a27605626528a3b2ce5563ae.

danog:
Thank you!

